PR #176, opened 2021-03-24
OPSEXP-849: Update generated README.txt to point to official docs rather than GitHub

nilsSure:
The generated README.txt file now points to docs.alfresco.com pages.
The text:
```
You can find full documentation on how to use the playbook for Enterprise deployments here https://docs.alfresco.com/content-services/latest/install/ansible and for Community deployments here https://docs.alfresco.com/content-services/community/install/ansible
```


gavincornwell (approved):
LGTM

PR #178, opened 2021-03-31
Opsexp 850

lsuciu:


nilsSure (approved):
LGTM

group_vars/all.yml
```diff
@@ -134,3 +134,6 @@ sync_db_driver: org.postgresql.Driver
 sync_db_name: "alfresco-sync"
 sync_db_username: "alfresco"
 sync_db_password: "alfresco"
+
+enable_ssl: false
```
gavincornwell, line 138: Just a naming comment, could we please stick to the convention we started with "use_custom_keystores" i.e. prefixing any feature flags with "use_" e.g. "use_ssl"

gavincornwell (changes requested):
As well as the automated test we discussed in the standup should we also be updating other properties, for example, shouldn't "repo.host" use the FQDN if there is one?

Or will you be doing that as part of https://alfresco.atlassian.net/browse/OPSEXP-650 ?

PR #180, opened 2021-04-01
OPSEXP-813: Firewalld issues on clean machines

sergiuv2020:


nilsSure (approved):
Looks good.

docs/deployment-guide.md
```diff
@@ -385,6 +385,12 @@ To check your inventory file is configured correctly and the control node is abl
 ansible all -m ping -i inventory_ssh.yml
 ```
 
+**Optional** To check if the required ports for the deployment are available on the target machine and we also have connectivity between nodes (ex. repository connecting to the db on 5432) please run the prerequisite_checks playbook before you deploy ACS. If there are any firewals blocking connectivity this playbook will discover them.
```
gavincornwell, line 388: typo: "firewals" -> "firewalls"

gavincornwell (changes requested):
One minor typo in the docs, other than that LGTM.

Also, is there any way to incorporate this into the multi-machine integration test?

playbooks/prerequisite_checks.yml
```diff
@@ -0,0 +1,139 @@
+---
+- name: prerequisite checks
+  hosts: repository
+  become: true
+  become_method: sudo
```
alxgomz, line 5: sudo is anyway the default become_method and specifying it would make it impossible to run the prereq playbook on systems using su e.g.
As it is now I can't see any other reason other become_method wouldn't work.

alxgomz (changes requested):
It's pretty nice to have that!
Just one comment about the become_method.
Out of curiosity, why did you choose to add the module in a new role instead of the playbook itself?

playbooks/prerequisite_checks.yml
sergiuv2020: good chatch, removed the method in both playbooks

nilsSure (approved):
LGTM

PR #181, opened 2021-04-08
OPSEXP-865: port defatultvars

alxgomz:
move hardcoded port numbers to common/defaults/main.yml and make them variables (which can be set later down the road) and default to numeric values when not set.
tests have not been updated because:
* I couldn't find a way to tell the molecule[docker] driver to use a variable in its config (`exposed_ports`)
* tests are anyway connecting to the `published_ports` which is unrelated to the ports pods actually listen on, and are set statically in the molecule[docker] `platform` config


sergiuv2020 (commented):
Could you please also remove the hardcodings in the prerequisites checks. It should be ok to just import the default vars from common before everything.

gavincornwell (commented):
Nothing more to add that hasn't been said already, looks good.

nilsSure (approved):
LGTM

alxgomz:
> Could you please also remove the hardcodings in the prerequisites checks. It should be ok to just import the default vars from common before everything.

Done. I also had to add a retry mecanism as I had issues with the check mecanism when consecutives tasks test the same port from different hosts.

PR #183, opened 2021-04-12
OPSEXP-809: Upgrade DTAS usage to v1.0

gavincornwell:
DTAS_VERSION environment variable has been added to Travis build settings.

Fixed config files to use AMP id rather than name/title.


nilsSure (approved):
LGTM

PR #184, opened 2021-04-13
OPSEXP-881 Use ansible (local) lookups to prevent side effects of become tasks

alxgomz:
At the moment we're using a first task to test presence of a var file to include, and another one to do the actual include.
This is causing issues as described in OPSEXP-881
Using lookup instead of tasks allow to avoid being subject to become issue and is forced to be executed locally as described in ansible doc:
> [Like all templating, lookups execute and are evaluated on the Ansible control machine](https://docs.ansible.com/ansible/latest/user_guide/playbooks_lookups.html)

using `errors='ignore'` provide the desired behaviour initially described in OPSEXP-848

using loop over first_found(os_fallback) allow for:
 * fallback to major release distribution, distribution, parent distribution, system type in case distribution specific file is not defined
 * silently skip in if no match


gavincornwell (commented):
Looks OK to me but I'll leave approval to the experts ;-)

mteodori:
@alxgomz can you please add some more details in title and description?

nilsSure (approved):
LGTM

PR #185, opened 2021-04-14
OPSEXP-886 do not rely on releasever as its behaviour varies

alxgomz:
use appropriate ansible variable instead of $releasever


nilsSure (approved):
LGTM

PR #187, opened 2021-05-03
OPSEXP-887: ubuntu20 ac7ent

alxgomz:
* make tasks distro agnotisc as much as possible
* split distro specific tasks to separate files
* try to have a configuration tasks distro agnostic
* make 3rd party download distro agnostic
* create a new integration test for ubuntu (localhost deployment)
* fix some molecule tests


roles/postgres/tasks/apt_repo.yml
```diff
@@ -0,0 +1,16 @@
+---
+# Ansible module apt_repository rely on unmaintained python libs so using template instead
+- name: Install PostgreSQL deb repository
+  become: true
+  template:
+    src: apt_repo.j2
```
nilsSure, line 6: Why not use apt_repository?
https://docs.ansible.com/ansible/2.9/modules/apt_repository_module.html

roles/postgres/templates/apt_repo.j2
```diff
@@ -0,0 +1,3 @@
+# PostgreSQL APT repository (Installed by alfresco-ansible-deployment)
+#
+deb http://apt.postgresql.org/pub/repos/apt {{ ansible_lsb.codename }}-pgdg main
```
nilsSure, line 3: see my previous comment

roles/sfs/templates/alfresco-shared-fs.service
```diff
@@ -4,8 +4,8 @@ After=syslog.socket network.target
 
 [Service]
 Type=simple
-User=alfresco
-ExecStart=/opt/alfresco/ats-shared-fs.sh
+User={{ username }}
```
nilsSure, line 7: good catch :)

roles/transformers/defaults/main.yml
```diff
@@ -6,3 +6,11 @@ utils_transform:
...
   - unzip
   - which
+  - fontconfig
+  - dejavu-fonts-common
+  - fontpackages-filesystem
+  - freetype
+  - libpng
+  - dejavu-sans-fonts
+
+pkg_fmt: "{% if ansible_pkg_mgr == 'apt' %}deb{% else %}rpm{% endif %}"
```
nilsSure, line 16: this var not is not very descriptive :(

roles/transformers/tasks/deb-packages.yml
```diff
@@ -0,0 +1,14 @@
+---
```
nilsSure, line 1: These file names are not very consistent. Eg: yum_repo.yml 
At least for tasks we should be consistent with the naming (either use OS related tasks or package manager related tasks, etc.)

roles/transformers/tasks/main.yml
```diff
@@ -13,59 +35,18 @@
...
-    dest: "{{ download_location }}/libreoffice.tar.gz"
-    url_username: "{{ nexus_user }}"
-    url_password: "{{ nexus_password }}"
-    mode: 0444
-    timeout: 950
-  register: libreoffice_download
-  async: 900
-  poll: 0
+- name: Prepare RPM based systems
+  when: pkg_fmt != 'deb'
```
nilsSure, line 39: when condition is at the end of blocks throughout the code. weird placement :)

roles/transformers/tasks/main.yml
```diff
@@ -13,59 +35,18 @@
...
+  block:
+    - name: Enable EPEL repository
+      yum_repository:
```
nilsSure, line 42: I've seen that in other places you've handled dnf also. why not here?

roles/transformers/vars/main.yml
```diff
@@ -1,17 +1,17 @@
...
-imagemagick_sha1_checksum: "{{ lookup('url', 'https://nexus.alfresco.com/nexus/service/local/repositories/thirdparty/content/org/imagemagick/imagemagick-distribution/{{ imagemagick_version }}/imagemagick-distribution-{{ imagemagick_version }}-linux.rpm.sha1') }}"
-imagemagick_lib_rpm: "https://nexus.alfresco.com/nexus/service/local/repositories/thirdparty/content/org/imagemagick/imagemagick-distribution/{{ imagemagick_version }}/imagemagick-distribution-{{ imagemagick_version }}-libs-linux.rpm"
-imagemagick_lib_sha1_checksum: "{{ lookup('url', 'https://nexus.alfresco.com/nexus/service/local/repositories/thirdparty/content/org/imagemagick/imagemagick-distribution/{{ imagemagick_version }}/imagemagick-distribution-{{ imagemagick_version }}-libs-linux.rpm.sha1') }}"
-imagemagick_dep_rpm: "https://dl.fedoraproject.org/pub/epel/epel-release-latest-{{ ansible_distribution_major_version }}.noarch.rpm"
+libreoffice_repo: https://downloadarchive.documentfoundation.org/libreoffice/old
+libreoffice_archive: >-
+  {{ libreoffice_repo }}/{{ libreoffice }}/{%
```
nilsSure, line 5: this block is really hard to read :(

sergiuv2020 (approved):
looks ok but i still want to see the results from the comments me and Andrei added together.

roles/postgres/tasks/main.yml
```diff
@@ -3,36 +3,21 @@
...
-
-- name: Disable the built-in PostgreSQL module
-  command: dnf -qy module disable postgresql
-  args:
-    warn: false  # set warn=false to prevent warning
-  register: disable_dnf_module
-  changed_when: "'Nothing to do.' not in disable_dnf_module.stdout"
-  when: ansible_distribution_major_version == "8"
+- name: Prepare package manager
+  include_tasks: "{{ ansible_pkg_mgr }}_repo.yml"
```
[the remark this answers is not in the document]
alxgomz: because the included tasks depends on the package managers more than the distro itself.

roles/postgres/tasks/apt_repo.yml
alxgomz: Actually I tried to but failed due to some dependencies that were not playing well with python3.
IIRC the main problem I had was with python-distutils-extra

roles/transformers/tasks/deb-packages.yml
alxgomz: The reason I've done this is to avoid code duplication: having one file for yum and one for dnf that are identical

roles/transformers/tasks/main.yml
alxgomz: essentially because dnf just wraps yum. I think places where dnf is handled specifically are tasks that are specific to dnf?

roles/nginx/tasks/main.yml
```diff
@@ -1,27 +1,32 @@
...
+      retries: 20
+      delay: 6
+      until: yum_lock.stdout_lines|length <= 2
+      changed_when: false
+  when: ansible_pkg_mgr != 'apt'
 
 - name: Ensure nginx is installed.
   package:
-    name: "{{ nginx_pack }}"
+    name: "{{ nginx_pack }}{%if ansible_pkg_mgr != 'apt' %}-{{ nginx_version }}{% endif %}"
```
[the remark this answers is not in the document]
alxgomz: I think I did that in order to be conservative regarding what was done for centos. On Ubuntu at least, there is no reason to use on version of  nginx over another.
Though I've aligned both to behave the same

nilsSure (approved):
LGTM

sergiuv2020 (approved):

PR #188, opened 2021-05-13
OPSEXP-903: Update tomcat and java

nilsSure:
Updated tomcat to 8.5.65 and 9.0.45
Updated Java to  11.0.11


sergiuv2020 (approved):

roles/java/vars/main.yml
```diff
@@ -1,5 +1,5 @@
 ---
 # vars file for roles/java
 jdk_tar_file: "OpenJDK11U-jdk_x64_linux_{{ dependencies_version.jdk }}_{{ dependencies_version.jdk_arch }}.tar.gz"
-jdk_folder: "openjdk-{{ dependencies_version.jdk }}+{{ dependencies_version.jdk_arch }}"
+jdk_folder: "openjdk-{{ dependencies_version.jdk }}_{{ dependencies_version.jdk_arch }}"
```
alxgomz, line 4: this is worrying me as it may means that playbook fails if user points to a version where the content of the archive has + or any unexpected character...
Do we have any idea how much this change from openjdk is stable?

alxgomz (approved):
Otherwise LGTM

PR #210, opened 2021-11-23
Opsexp 651: Automate version update of deployment components in the readme

Mieszko96:
Scenario
Someone made changes in values
Someone want to commit chase
Pre-commit will fail (Readme updated by script) ﻿
Someone need to “git add .” ﻿
Someone commit again(with automated updated Readme file)

Tested, if new file 8.0.N-extra-vars.yml added to repository, will be included into table.

Problems:
developers using this repository need to have installed python3 with pyyaml library.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,3 +34,10 @@ repos:
     hooks:
       - id: giticket
         args: ['--regex', '[A-Z]+-\d+[0-9]', '--format', '{ticket}: {commit_msg}', '--mode', 'regex_match']
+
+  - repo: local
+    hooks:
+      - id: custom
+        name: Generate-components-version-table
+        entry: scripts/generate-comp-ver-table.py
+        language: python
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -38,6 +38,6 @@ repos:
   - repo: local
     hooks:
       - id: custom
-        name: Generate-components-version-table
+        name: Generate components version table
         entry: scripts/generate-comp-ver-table.py
         language: python
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,7 +37,7 @@ repos:
 
   - repo: local
     hooks:
-      - id: custom
+      - id: generate-table
         name: Generate components version table
         entry: scripts/generate-comp-ver-table.py
         language: python
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,10 +40,10 @@ jobs:
       script:
         - yamllint .
     - name: Check components version table
-      cripts:
+      scripts:
         - python3 scripts/generate-comp-ver-table.py
         - git_status_out=$(git status)
-        - if [[ "$git_status_out" == *"docs/README.md"* ]]; then exit 1; fi
+        - if [[ "$git_status_out" == *"docs/README.md"* ]]; then echo "Error- Components version was not changed. Download pre-commit tool";exit 1; fi
     - stage: Verification
       name: Search Role Verification
       script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,11 +39,6 @@ jobs:
     - name: "yamllint"
       script:
         - yamllint .
-    - name: Check components version table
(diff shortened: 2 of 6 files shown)


.pre-commit-config.yaml
```diff
@@ -34,3 +34,10 @@ repos:
     hooks:
       - id: giticket
         args: ['--regex', '[A-Z]+-\d+[0-9]', '--format', '{ticket}: {commit_msg}', '--mode', 'regex_match']
+
+  - repo: local
+    hooks:
+      - id: custom
```
alxgomz, line 40: I would use a more descriptive id is possible

scripts/generate-comp-ver-table.py
```diff
@@ -0,0 +1,132 @@
...
+            line_in_table+=f"{version} | "
+
+        line_in_table=line_in_table[:-1]
+        end_table+=f"{line_in_table}\n"
+    return end_table
+
+def modify_table():
+    """Delete old table from README.md, and put new one"""
+    delete_next_lines = False
+    a_file = open("docs/README.md", "r", encoding="utf-8")
```
alxgomz, line 112: Does this pass pylint? I thought with syntax was now the way to go with python3

scripts/generate-comp-ver-table.py
```diff
@@ -0,0 +1,132 @@
...
+        if "| Sync Service | " in line.strip("\n"):
+            delete_next_lines = False
+            new_file.write(get_content_of_new_table())
+        elif delete_next_lines is True:
+            pass
+        elif "| Component | " in line.strip("\n"):
+            delete_next_lines = True
+        else:
+            new_file.write(line)
+
```
alxgomz, line 128: That logic relies heavily on the fact the table will always keep the same structure, ordering, labelling and so on...
I'm afraid this makes it somewhat fragile.
Did you investigate whether a markdown library would have allowed table manipulation in a more solid manner? (Just wondering, I have not done any investigation myself)

alxgomz (changes requested):
Only the later comment is a real concern

scripts/generate-comp-ver-table.py
Mieszko96: I will make investigation

scripts/generate-comp-ver-table.py
Mieszko96: Passed pylint, i can change to "with open"

.pre-commit-config.yaml
Mieszko96: Will change

scripts/generate-comp-ver-table.py
Mieszko96: One idea is to create this table in seperate file. In README.md will be reference to this file.

alxgomz: as agreed with @mteodori it's ok to keep it like that

alxgomz (approved):
LGTM

alxgomz (changes requested):
Can you please add pyyaml to the `requirements.txt` file and add that check to the Linters stage (in case dev doesn't have pre-commit installed locally?

Mieszko96:
> Can you please add pyyaml to the `requirements.txt` file and add that check to the Linters stage (in case dev doesn't have pre-commit installed locally?

.travis.yml
```diff
@@ -39,6 +39,11 @@ jobs:
     - name: "yamllint"
       script:
         - yamllint .
+    - name: Check components version table
+      cripts:
```
alxgomz, line 43: looks like a typo

scripts/generate-comp-ver-table.py
```diff
@@ -0,0 +1,129 @@
...
+"""Generate components version table in docs/README.md based on extra-vars.yml."""
+
+import sys
+import glob
+import yaml
+
+
+def get_acs_extra_values_files():
+    """Return list of extra-values files in repository."""
```
alxgomz, line 11: you still have an extra-values instead of extra-vars here

alxgomz (approved):
LGTM

PR #216, opened 2021-11-30
OPSEXP-1145: Fix ansible deployment using python 2

Mieszko96:


.travis.yml
```diff
@@ -344,6 +344,19 @@ jobs:
         - echo "wait for repo to be up" && sleep 5m
         - ssh ec2-user@$control_machine_ip "$COMMON_COMMANDS; sed -i 's+TEST_URL+http://$TESTHOST+g' tests/test-config.json; cd dtas; pytest --tb=line --configuration ../tests/test-config.json tests/ -s"
         - if [[ "$TRAVIS_COMMIT_MESSAGE" != *"[keep env]"* ]]; then ssh -tt ec2-user@$control_machine_ip "$COMMON_COMMANDS;molecule destroy -s ec2multi"; fi
+    - stage: Cleanup
```
alxgomz, line 347: Can you please make it an `after_script`?
This is only ever useful for the  single & multi-machine stages. We don't need to boot up another VM for a new stage and you would also avoid spreading the dodgy "preliminary code". Also each stage can delete its own work folder and avoid mess up with another stage's dir

alxgomz (changes requested):
well done finding all occurrence of that code I missed them!

alxgomz (approved):
LGTM

PR #218, opened 2021-12-01
OPSEXP-1154 fix string split syntax

alxgomz:
Fix problems with alfresco ImageMagick (aio) tengine.
Ref: OPSEXP-1154


Mieszko96 (approved):
LGTM

Mieszko96 (approved):
+2

Issue #219, opened 2021-12-07
.sh files with setenv path hard coded


luisfernandezprado:
Hi,

It is detected that the following "sh" boot files have the path of the "setenv.sh" hard coded without using the ansible configuration path {{ config_folder }}.

The following change is proposed
```bash
## original
. /etc/opt/alfresco/setenv.sh
## change
. {{ config_folder }}/setenv.sh
```

Affected files
- alfresco-ansible-deployment\roles\sfs\templates\ats-shared-fs.sh
- alfresco-ansible-deployment\roles\sync\templates\syncservice.sh
- alfresco-ansible-deployment\roles\transformers\templates\ats-ate-aio.sh
- alfresco-ansible-deployment\roles\trouter\templates\ats-atr.sh

howkymike:
Well, I also lost a few seconds to spot this mistake (and also wanted to PR this). I wonder why is it so complicated and long-lasting to fix such a small bug.

gionn:
@luisfernandezprado @howkymike thanks for reporting! All of your concerns should have been addressed, plus some more.

PR #220, opened 2021-12-07
Code to resolve the issue #219

luisfernandezprado:
https://github.com/Alfresco/alfresco-ansible-deployment/issues/219

It is detected that the following "sh" boot files have the path of the "setenv.sh" hard coded without using the ansible configuration path {{ config_folder }}.

The following change is proposed

```bash
## original
. /etc/opt/alfresco/setenv.sh
## change
. {{ config_folder }}/setenv.sh
```

Affected files

- alfresco-ansible-deployment\roles\sfs\templates\ats-shared-fs.sh

- alfresco-ansible-deployment\roles\sync\templates\syncservice.sh

- alfresco-ansible-deployment\roles\transformers\templates\ats-ate-aio.sh

- alfresco-ansible-deployment\roles\trouter\templates\ats-atr.sh

### DIFF
--- a/roles/sfs/templates/ats-shared-fs.sh
+++ b/roles/sfs/templates/ats-shared-fs.sh
@@ -4,7 +4,7 @@ if [ $(id -u) -eq 0 ]; then
     exit
 fi
 
-. /etc/opt/alfresco/setenv.sh
+. {{ config_folder }}/setenv.sh
 {% for key, value in sfs_environment.items() %}
 export {{key}}="{{value}}"
 {% endfor %}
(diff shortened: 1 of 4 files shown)


luisfernandezprado:
I am not sure how to proceed to sign the license.... I have entered the "license/cla" page and it tells me that I have already signed... but even so the message does not go away.

I don't know if I have to do any more steps

image

image

alxgomz:
Hi @luisfernandezprado 
Good catch, and thank for that PR which looks OK.

However, in order to make sure all code gets tested properly we ask you to raise pull requests against the `contribs` branch instead of using `master`, or any other branch.
Please raise that same PR again but against `contribs`

PR #227, opened 2022-01-06
OPSEXP-1131 ACS hotfix upgrade

alxgomz:
Allow Ansible playbook to proceed to "in-place" upgrades for hotfixes
Previous tasks were using "lineinfile" module which can be tricky to use for formats like xml and idempotency.
Replaced by templates

Ref. OPSEXP-1131

### DIFF
--- a/docs/components-upgrade.md
+++ b/docs/components-upgrade.md
@@ -59,6 +59,9 @@ acs:
 ...
 ```
 
+> IMPORTANT: make sure you do not set the version to a version number that's not a hotfix (version number needs to be 4 digits and the 3 first ones needs to match the ones of the initially deployed version)
+> This is because, as eplained earlier, "in-place" upgrades are only supported for hotfixes
+
 Once these changes are saved run the command bellow:
 
 ```
--- a/docs/components-upgrade.md
+++ b/docs/components-upgrade.md
@@ -37,9 +37,9 @@ Alfresco also releases some hotfixes and an hotfix upgrade would be moving from 
 
 In order to apply a later hotfix, you need to first match the pre-requisites, then change the ACS version to point to the hotfix version in the appropriate file, and finally run the playbook again.
 
-In the example bellow we want to upgrade from the initial 7.0.1 installtion to 7.0.1.4 hotfix:
+In the example bellow we want to upgrade from the initial 7.0.1 installation to 7.0.1.4 hotfix:
 
-Edit `7.0.N-extra-vars.yml` and changes the the bellow snippet
+Edit `7.0.N-extra-vars.yml` and changes the the bellow snippet:
 
 ```
 ---           
@@ -49,7 +49,7 @@ acs:
 ...
 ```
 
-to
+to:
 
 ```
 ---           
--- a/docs/components-upgrade.md
+++ b/docs/components-upgrade.md
@@ -1,0 +1,77 @@
+# Upgrading an environment
+
+The ACS installation is made of several components among which:
+
+ - Content service
+ - share
+ - digital workspace
+ - Search service
+ - ...
+
+## Pre-requisites
+
+Before proceeding to an upgrade, an administrator needs to:
+
+ * Make sure the [upgrade path and documentation](https://docs.alfresco.com/content-services/latest/upgrade/) is followed
+
+ * Appropriate backups of database, contentstore and indexes are done and can be restored in case a rollback is needed
+
+ * Initial deployment must have been done using this Ansible playbook
+
+## Supported Components
+
+### Content service
+
+#### Upgrade types
+
+There are different kinds of upgrade one may want to proceed. For example a major upgrade would be upgrading from 7.0.x to 7.1.x.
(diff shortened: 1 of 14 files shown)


Mieszko96 (approved):
LGTM

docs/components-upgrade.md
```diff
@@ -0,0 +1,80 @@
...
+
+Alfresco also releases some hotfixes and an hotfix upgrade would be moving from 7.0.1 to 7.0.1.4. This type of upgrade can be done "in-place" and is supported by the playbook.
+
+> Please note that "in-place" upgrade still need to match the upgrade pre-requisites
+
+#### Proceeding to an "in-place" hotfix upgrade
+
+In order to apply a later hotfix, you need to first match the pre-requisites, then change the ACS version to point to the hotfix version in the appropriate file, and finally run the playbook again.
+
+In the example bellow we want to upgrade from the initial 7.0.1 installtion to 7.0.1.4 hotfix:
```
gionn, line 40:
```suggestion
In the example bellow we want to upgrade from the initial 7.0.1 installation to 7.0.1.4 hotfix:
```

docs/components-upgrade.md
```diff
@@ -0,0 +1,80 @@
...
+
+Edit `7.0.N-extra-vars.yml` and changes the the bellow snippet
```
gionn, line 42:
```suggestion
Edit `7.0.N-extra-vars.yml` and changes the the bellow snippet:
```

docs/components-upgrade.md
```diff
@@ -0,0 +1,80 @@
...
+
+```
+---           
+acs:         
+  version: 7.0.1
+  edition: Enterprise 
+...
+```
+
+to
```
gionn, line 52:
```suggestion
to:
```

docs/components-upgrade.md
gionn: for consistency

docs/components-upgrade.md
```diff
@@ -0,0 +1,80 @@
...
+```
+---           
+acs:         
+  version: 7.0.1.4
+  edition: Enterprise 
+...
+```
+
+> IMPORTANT: make sure you do not set the version to a version number that's not a hotfix (version number needs to be 4 digits and the 3 first ones needs to match the ones of the initially deployed version)
+> This is because, as eplained earlier, "in-place" upgrades are only supported for hotfixes
```
gionn, line 63:
```suggestion
> This is because, as explained earlier, "in-place" upgrades are only supported for hotfixes
```

docs/components-upgrade.md
```diff
@@ -0,0 +1,80 @@
...
+
+```
+---           
+acs:         
+  version: 7.0.1.4
+  edition: Enterprise 
+...
+```
+
+> IMPORTANT: make sure you do not set the version to a version number that's not a hotfix (version number needs to be 4 digits and the 3 first ones needs to match the ones of the initially deployed version)
```
gionn, line 62: what are the side effects if someone doesn't read in time this important note?

Maybe isn't trivial to check semantically that the new version is not an in-place upgrade, but I am asking just to be sure that for any chance this has been already discussed elsewhere

alxgomz: That's a valid concern. People could end up trying an upgrade that's not supported.
E.g. if they try a service pack upgrade, that may go well, but in case of major upgrade chances are that would break the installation.
Thinking of it maybe we could add a file on each target host which describe which version of the components have been deployed locally on the host together with an additional setting to allow for upgrades... That would allow to:
* make sure people know what they are doing
* check the version match what was previously installed

What do you think?

gionn: I would implement it, even in most simple way, to avoid unintended disruptions (if versionInstalled != desiredVersion throw error), the override can just be to manually update the guard file with the desiredVersion (don't know if it make sense with Ansible, maybe in a couple of months I will understand it lol)

roles/common/tasks/check_upgrades.yml
```diff
@@ -0,0 +1,73 @@
...
+        - installed_components.sfs is defined
+        - sfs != installed_components.sfs
+          
+    - name: Fail on unsupported Search  upgrades
+      fail:
+      when:
+          - installed_components.search is defined
+          - search != installed_components.search
+
+    - name: Fail on unsupported Sync  upgrades
```
gionn, line 61:
```suggestion
    - name: Fail on unsupported Sync upgrades
```

roles/common/tasks/check_upgrades.yml
```diff
@@ -0,0 +1,73 @@
...
+        - installed_components.trouter is defined
+        - trouter != installed_components.trouter
+
+    - name: Fail on unsupported SFS upgrades
+      fail:
+      when:
+        - installed_components.sfs is defined
+        - sfs != installed_components.sfs
+          
+    - name: Fail on unsupported Search  upgrades
```
gionn, line 55:
```suggestion
    - name: Fail on unsupported Search upgrades
```

PR #228, opened 2022-01-11
OPSEXP-1172 fix hardcoded templates

### LINKED ISSUE #219: .sh files with setenv path hard coded

gionn:
ref OPSEXP-1112
fix #219

### DIFF
--- a/roles/activemq/tasks/main.yml
+++ b/roles/activemq/tasks/main.yml
@@ -33,7 +33,7 @@
     remote_src: true
     owner: "{{ username }}"
     group: "{{ group_name }}"
-    creates: /opt/apache-activemq-{{ dependencies_version.activemq }}/bin/activemq
+    creates: "{{ activemq_home }}/bin/activemq"
   when: job_result is succeeded
 
 - name: Move configuration files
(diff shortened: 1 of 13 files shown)


alxgomz:
we also have a bunch of pytest script which still have those hardcoded values:
* ./roles/repository/molecule/default/tests/test_repository.py
* ./roles/sfs/molecule/default/tests/test_sfs.py
* ./roles/trouter/molecule/default/tests/test_trouter.py
and. also a molecule playbook:
* ./roles/repository/molecule/default/converge.yml
and maybe others

alxgomz (changes requested):
there are more occurrences of these hardcoded strings

gionn:
> we also have a bunch of pytest script which still have those hardcoded values:

it may be better to keep tests unlinked from defaults defined in roles, I would expect tests failure after changing a variable containing a path rather than all green tests.

alxgomz (approved):
LGTM

PR #230, opened 2022-01-24
OPSEXP-1054 Enable dependabot for pip dependencies

gionn:
OPSEXP-1054


alxgomz (approved):
LGTM (but we might relax that in the future... in particular for linters)

PR #233, opened 2022-01-26
OPSEXP-1197 Refactor localhost integration stage config

gionn:
OPSEXP-1197


gionn:
I will try to migrate these stages to be driven by molecule only and remove the entire custom code handling in master

PR #234, opened 2022-01-26
OPSEXP-1057 Switch to official centos:8 image for molecule tests

gionn:
OPSEXP-1057


configuration_files/keystores/.gitignore
```diff
@@ -0,0 +1 @@
+*
```
alxgomz, line 1: Can we get that centralized in the root `.gitignore`?

alxgomz (changes requested):

alxgomz (approved):
LGTM

PR #242, opened 2022-02-03
OPSEXP-1056 Test roles on ubuntu

gionn:
OPSEXP-1056


.travis.yml
```diff
@@ -48,335 +48,213 @@ jobs:
...
       script:
-        - cd roles/sync
+        - cd roles/transformers
         - molecule converge
         - sleep 1m
         - molecule verify
         - molecule destroy
+
     - stage: Verification
-      name: Localhost integration checks (ACS 6.2.2 on CentOS 7)
```
gionn, line 121: to revert before merge

.travis.yml
```diff
@@ -26,6 +26,7 @@ env:
     - ANSIBLE_VERSION="2.9.21"
     - AWS_REGION=us-east-1
     - DTAS_VERSION=v1.1.1
+    - "MOLECULE_ROLE_IMAGE=ubuntu:20.04"
```
gionn, line 29: to be removed

.travis.yml
```diff
@@ -70,37 +70,79 @@ jobs:
...
-        - molecule --version
-        - cd roles/search/
-        - molecule converge
-        - sleep 1m
-        - molecule verify
-        - molecule destroy
+        - cd roles/search && molecule test
 
     - stage: Roles tests
-      name: Common Role Verification
```
alxgomz, line 83: Given all other roles are dependant on the common role and we doubled the number of tests I think we can remove testing this specific role

.travis.yml
```diff
@@ -70,37 +70,79 @@ jobs:
...
+
+    - stage: Roles tests
+      name: Java tests on Ubuntu
+      env:
+        - "MOLECULE_ROLE_IMAGE=ubuntu:20.04"
       script:
         - cd roles/java/ && molecule test
 
     - stage: Roles tests
-      name: Tomcat Role Verification
```
alxgomz, line 93: Given the repo role is dependant on the tomcat role and we doubled the number of tests I think we can remove testing this specific role

roles/activemq/molecule/default/molecule.yml
```diff
@@ -1,45 +1,20 @@
...
-      - 5432/tcp
-      - 8161/tcp
-      - 8080/tcp
-      - 8082/tcp
-      - 8083/tcp
-      - 8090/tcp
-      - 8099/tcp
-      - 8983/tcp
-      - 9090/tcp
     published_ports:
```
alxgomz, line 18: I don't understand why we need all these ports published for activemq tests. Actually the activemq port itself is not even part of the list, and as we use the docker exec driver to test on localhost... I suspect neither of theese ports are really needed. Worth checking what's really needed here

roles/adw/molecule/default/molecule.yml
```diff
@@ -1,47 +1,20 @@
...
-      - 8080/tcp
-      - 8082/tcp
-      - 8083/tcp
-      - 8090/tcp
-      - 8099/tcp
-      - 8880/tcp
-      - 8888/tcp
-      - 8983/tcp
-      - 9090/tcp
     published_ports:
```
alxgomz, line 18: same as above: which ports do we really needed here?

roles/molecule/Dockerfile.j2
```diff
@@ -0,0 +1,22 @@
+# Forked from https://github.com/ansible-community/molecule-docker/blob/main/src/molecule_docker/playbooks/Dockerfile.j2
```
alxgomz, line 1: having this folder suggests we have a `molecule` role which is not true. I'd rather we use the root `/molecule` folder for that

roles/sync/molecule/default/molecule.yml
```diff
@@ -1,44 +1,20 @@
...
-      - 5432/tcp
-      - 8161/tcp
-      - 8080/tcp
-      - 8082/tcp
-      - 8083/tcp
-      - 8090/tcp
-      - 8099/tcp
-      - 8983/tcp
-      - 9090/tcp
     published_ports:
```
alxgomz, line 18: same comments about ports here

roles/tomcat/molecule/default/molecule.yml
```diff
@@ -1,45 +1,20 @@
...
-      - 5432/tcp
-      - 8161/tcp
-      - 8080/tcp
-      - 8082/tcp
-      - 8083/tcp
-      - 8090/tcp
-      - 8099/tcp
-      - 8983/tcp
-      - 9090/tcp
     published_ports:
```
alxgomz, line 18: same comment about ports here

roles/transformers/molecule/default/molecule.yml
```diff
@@ -1,46 +1,20 @@
...
-      - 8161/tcp
-      - 8080/tcp
-      - 8082/tcp
-      - 8083/tcp
-      - 8090/tcp
-      - 8099/tcp
-      - 8983/tcp
-      - 9090/tcp
-      - 61616/tcp
     published_ports:
```
alxgomz, line 18: same comment about ports here

alxgomz (changes requested):
only real concerns is about the location of the Dockerfile template. Other comments are just more cleaning stuffs.

.travis.yml
gionn: in that way we are losing the tests already written for that particular role

Can make sense to avoid running suites on multiple platforms that are going to be tested in another suite, so I will leave only one run for common, java, tomcat roles (I would prefer ubuntu 20.04 that is newer than centos:7)

alxgomz: make sense

roles/tomcat/molecule/default/molecule.yml
```diff
@@ -1,75 +1,32 @@
...
-  directory: tests
   env:
     TEST_HOST: "localhost"
-    ANSIBLE_TIMEOUT: "30"
-    ANSIBLE_SSH_RETRIES: "15"
-    PYTHONWARNINGS: "ignore:.*U.*mode is deprecated:DeprecationWarning"
   options:
-    v: 1
+    verbose: true
     s: true
```
alxgomz, line 32: Did you intend to keep that verifier switch in the tomcat tests while it's been removed from all others?

alxgomz (approved):
LGTM (please just review my comment)

roles/tomcat/molecule/default/molecule.yml
gionn: no it's only a leftover, I didn't have found a good reason to explicitly disable i/o capture and I get rid of them

verbose is useful because tests looks stuck until the end and was confusing me :D

PR #245, opened 2022-02-10
OPSEXP-1232 multimachine ec2

gionn:
OPSEXP-1232


molecule/default/create.yml
```diff
@@ -38,6 +38,14 @@
...
         to_port: "{{ default_ssh_port }}"
         cidr_ip: "0.0.0.0/0"
+      - proto: tcp
+        from_port: 443
+        to_port: 443
+        cidr_ip: "0.0.0.0/0"
+      - proto: tcp
+        from_port: "0"
+        to_port: "65535"
+        group_name: "{{ default_security_group_name }}"
```
gionn, line 48: moved back to the default security group in the create playbook because I didn't find an easy way to manage more than one security group on the nginx instance (tbh didn't want to spend more time on that one)

molecule/default/create.yml
```diff
@@ -38,6 +38,14 @@
         from_port: "{{ default_ssh_port }}"
         to_port: "{{ default_ssh_port }}"
         cidr_ip: "0.0.0.0/0"
+      - proto: tcp
+        from_port: 443
+        to_port: 443
+        cidr_ip: "0.0.0.0/0"
+      - proto: tcp
+        from_port: "0"
+        to_port: "65535"
```
alxgomz, line 47: Doesn't that work with molecule?

```suggestion
      - proto: all
```

molecule/default/verify.yml
```diff
@@ -1,6 +1,6 @@
 ---
 - name: Verify
-  hosts: all
+  hosts: nginx
```
alxgomz, line 3: You are setting target to hosts `nginx` but all tasks use `delegate_to: localhost`.
Why can't we set target hosts to `localhost` directly and remove tasks delegations?

alxgomz (changes requested):
It's sooo good to see a proper molecule generated inventory ❤️ 🚀 
Just minor stuffs

molecule/default/create.yml
gionn: it doesn't look like a supported value
https://docs.ansible.com/ansible/latest/collections/amazon/aws/ec2_group_module.html#parameter-rules/proto

alxgomz: It is, check in the examples on that same page. (I've also used it succesfully before)

molecule/default/verify.yml
```diff
@@ -9,30 +9,27 @@
   tasks:
 
     - name: Gather EC2 facts
+      delegate_to: "{{ groups['nginx'][0] }}"
```
alxgomz, line 12: I think this is useless as fact are only ever delegated if delegate_facts is set to true.
Also in the later tasks the ec2 facts you used are used from localhost (not delegated) tasks.

alxgomz (approved):
Just one comment I think it's ok without it anyway. Up to you.

PR #261, opened 2022-02-21
OPSEXP-1198 Make sure that root ebs is deleted upon termination for EC2 tests

gionn:
OPSEXP-1198

### DIFF
--- a/molecule/default/create.yml
+++ b/molecule/default/create.yml
@@ -48,6 +48,12 @@
         to_port: 0
         cidr_ip: "0.0.0.0/0"
 
+    default_root_volume:
+      device_name: /dev/sda1
+      ebs:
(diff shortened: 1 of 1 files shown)


molecule/default/create.yml
```diff
@@ -48,6 +48,12 @@
         to_port: 0
         cidr_ip: "0.0.0.0/0"
 
+    default_root_volume:
+      device_name: /dev/sda1
+      ebs:
+        volume_size: 10
+        delete_on_termination: yes
```
gionn, line 55: Screenshot 2022-02-21 at 12 44 36

alxgomz (approved):
LGTM (worth a PR in molecule-ec2?)

PR #251, opened 2022-02-14
OPSEXP-1250 avoid general failure with single failing checksum url

gionn:
OPSEXP-1250

PR #263, opened 2022-02-21
OPSEXP-1239 allow external contentstore

alxgomz:
Allow for customer to specify the contentstore  location so he can leverage his own architecture to provide HA of other enterprise features.


roles/repository/molecule/default/molecule.yml
```diff
@@ -4,7 +4,17 @@ dependency:
 driver:
   name: docker
 platforms:
-  - name: repository
+  - name: dummy
```
gionn, line 7: this is for having hostvars for these hosts available without really having to provision them, correct?

roles/repository/molecule/default/host_vars/repository1.yml
```diff
@@ -0,0 +1,51 @@
...
+    dest: "{{ content_folder }}/amps_repo/alfresco-aos-module.amp"
+  - url: "{{ nexus_repository.enterprise_releases }}/alfresco-governance-services-enterprise-repo/{{ amps.ags_repo.version }}/alfresco-governance-services-enterprise-repo-{{ amps.ags_repo.version }}.amp"
+    sha1_checksum: "{{ lookup('url', '{{ nexus_repository.enterprise_releases }}/alfresco-governance-services-enterprise-repo/{{ amps.ags_repo.version }}/alfresco-governance-services-enterprise-repo-{{ amps.ags_repo.version }}.amp.sha1', username=lookup('env', 'NEXUS_USERNAME'), password=lookup('env', 'NEXUS_PASSWORD')) }}"
+    dest: "{{ content_folder }}/amps_repo/ags-repo.amp"
+  - url: "{{ nexus_repository.enterprise_releases }}/alfresco-governance-services-enterprise-share/{{ amps.ags_share.version }}/alfresco-governance-services-enterprise-share-{{ amps.ags_share.version }}.amp"
+    sha1_checksum: "{{ lookup('url', '{{ nexus_repository.enterprise_releases }}/alfresco-governance-services-enterprise-share/{{ amps.ags_share.version }}/alfresco-governance-services-enterprise-share-{{ amps.ags_share.version }}.amp.sha1', username=lookup('env', 'NEXUS_USERNAME'), password=lookup('env', 'NEXUS_PASSWORD')) }}"
```
gionn, line 35: need to decide if applying here the same changes of https://github.com/Alfresco/alfresco-ansible-deployment/pull/251 or rework that after merging this PR (probably this pr has higher priority?)

roles/repository/molecule/default/molecule.yml
alxgomz: not really the hosts vars but more the structure of the inventory which we require. Basically I think we should do that everywhere in molecule config rather than linking the inventory file in ansible args

roles/repository/molecule/default/host_vars/repository1.yml
alxgomz: Ooch completely forgot about it!
Do you mind adding handling that change in your PR (I don't think one has higher prio given we anyway released 1.2.0 but I think it would make more sense...?

PR #269, opened 2022-02-24
OPSEXP-1237: Allow deploying repository and share cluster

alxgomz:
Make the playbook able to deploy a working cluster of repository and share nodes, and have all linked components leveraging the proxy to load balance requests over the different repo nodes


docs/deployment-guide.md
```diff
@@ -450,6 +450,36 @@ Once ACS has initialized access the system using the following URLs with a brows
...
+      hosts:
+        ecm1.infra.local:
+        ecm2.infra.local:
+        ingester.infra.local:
+          cluster_keepoff: true
+...
+```
+
+In some circumstances, you may want to have a repo node that's dedicated to a scheduled task (such as ingesting massive amount of documents). Depending on the nature of the task and the requirements of your organisation, it may be preferable to not make this node part of the ACS cluster.
+In that case, you can add the `cluster_keepoff` variable to one of the `repository` group nodes'. It will provision the node with teh repository and share services but make sure it not taking part in neither the share, nor the repository cluster realm.
```
gionn, line 479:
```suggestion
In that case, you can add the `cluster_keepoff` variable to one of the `repository` group nodes'. It will provision the node with the repository and share services but make sure it not taking part in neither the share, nor the repository cluster realm.
```

docs/shared-contentstore.md
```diff
@@ -85,3 +85,54 @@ all:
 ```
 
 > The example above could be used in case of sites replicated real-time through low latency links between sites. That's a feature high storage vendors can offer.
+
+## Dealing with permissions
+
+It is important to note that all of the repository cluster members needs to be able to write to the whole directory structure of the contentstore. Depending on your underlying storage technology this can have different implications.
+In many cases POSIX permission model applies. However there are 2 things to take into account re garding how the playbook works:
+
+- Installed services are run as a specific user `alfresco`, part of a specific group `Àlfresco`.
```
gionn, line 94:
```suggestion
- Installed services are run as a specific user `alfresco`, part of a specific group `alfresco`.
```

docs/shared-contentstore.md
```diff
@@ -85,3 +85,54 @@ all:
...
+
+> Keep in mind that, not only the root of the contentstore needs to be write-able by every node, but so does each folder created by any of the nodes.
+
+We'll describe bellow ways to fix or mitigate that problem.
+
+### Mitigations
+
+If all the targets repository hosts are built from the same OS image and strictly confugred the same way, there is no to very little chance for the UID/GID to get inconsistent accross nodes. This greatly limits the problem and most of the time if you stick to that principal you can use regular POSIX model approach
+
+If all the target repository hosts are bound to a centralized user/group directory, an option can be to manually created the `alfresco` user and `Alfresco` group within this directory so all nodes will for sure share the same UID/GID. Though you need to realize that it introduce a dependency on the platform to that directory system.
```
gionn, line 107:
```suggestion
If all the target repository hosts are bound to a centralized user/group directory, an option can be to manually created the `alfresco` user and `alfresco` group within this directory so all nodes will for sure share the same UID/GID. Though you need to realize that it introduce a dependency on the platform to that directory system.
```

docs/shared-contentstore.md
```diff
@@ -85,3 +85,54 @@ all:
...
+
+### Properly configuring your storage system
+
+Many storage technologies implement features to workaround this problem. Please see bellow some examples showing how to reliably setup shared storage for diverse clients.
+
+#### NFS
+
+This example takes advantage of the squashing mecanism NFS offers. this basically allows to map all NFS client requests to a single user (usually `nobody`) regardless of the UID/GID seen on the client.
```
gionn, line 115:
```suggestion
This example takes advantage of the squashing mechanism NFS offers. This basically allows to map all NFS client requests to a single user (usually `nobody`) regardless of the UID/GID seen on the client.
```

molecule/default/cleanup.yml
```diff
@@ -0,0 +1,32 @@
...
+  hosts: localhost
+  gather_facts: false
+  vars:
+    project_dir: "{{ lookup('env', 'MOLECULE_PROJECT_DIRECTORY') }}"
+  tasks:
+    - name: Gather EC2 facts
+      amazon.aws.ec2_metadata_facts:
+      delegate_to: "{{ groups['nginx'] | first }}"
+
+    - name: "Set variables"
```
gionn, line 12:
```suggestion
    - name: Set variables
```

molecule/default/cleanup.yml
```diff
@@ -0,0 +1,32 @@
...
+      set_fact:
+        certificate_domain: "{{ ansible_ec2_public_hostname }}"
+        private_key_path: "{{ project_dir }}/configuration_files/ssl_certificates/{{ ansible_ec2_public_hostname }}.key"
+        cert_path: "{{ project_dir }}/configuration_files/ssl_certificates/{{ ansible_ec2_public_hostname }}.crt"
+
+    - name: cleanup cert
```
gionn, line 18:
```suggestion
    - name: Cleanup cert
```

roles/common/defaults/main.yml
```diff
@@ -11,8 +11,9 @@ ats_home: "{{ binaries_folder }}/transform-service"
 
 # Due to https://github.com/ansible/ansible/issues/64957
 # Using default() filter for hostvars 1st level key requires ansible 2.10+
-nginx_host: "{% if groups.nginx | length == 0 %}127.0.0.1{% else %}{{ hostvars[groups.nginx[0]].ansible_host | default('127.0.0.1') }}{% endif %}"
-repo_host: "{% if groups.repository | length == 0 %}127.0.0.1{% else %}{{ hostvars[groups.repository[0]].ansible_default_ipv4.address | default('127.0.0.1') }}{% endif %}"
+nginx_host: "{{ groups['nginx'] | map('extract', hostvars, ['ansible_default_ipv4','address']) | first | default('127.0.0.1') }}"
+repo_hosts: "{{ groups['repository'] | map('extract', hostvars, ['ansible_default_ipv4','address']) | list }}"
+repo_host: "{% if repo_hosts %}{{ repo_hosts[0] }}{% else %}{{ ['127.0.0.1'] | list }}{% endif %}"
```
gionn, line 16: does it really make sense to provide a 127.0.0.1 default both here and for nginx_host? is there an use case?

roles/common/defaults/main.yml
```diff
@@ -11,8 +11,17 @@ ats_home: "{{ binaries_folder }}/transform-service"
 
 # Due to https://github.com/ansible/ansible/issues/64957
 # Using default() filter for hostvars 1st level key requires ansible 2.10+
-nginx_host: "{% if groups.nginx | length == 0 %}127.0.0.1{% else %}{{ hostvars[groups.nginx[0]].ansible_host | default('127.0.0.1') }}{% endif %}"
-repo_host: "{% if groups.repository | length == 0 %}127.0.0.1{% else %}{{ hostvars[groups.repository[0]].ansible_default_ipv4.address | default('127.0.0.1') }}{% endif %}"
+nginx_host: "{{ groups['nginx'] | map('extract', hostvars, ['ansible_default_ipv4','address']) | first | default('127.0.0.1') }}"
+repo_hosts: |
+  {{ groups['repository'] | map('extract', hostvars)
```
alxgomz, line 16: @gionn can you take a look at this and let me know what you think about using that approach for OPSEXP-1238?

roles/repository/templates/share-config-custom.xml
```diff
@@ -59,7 +59,7 @@
          </action>
       </rule>
    -->
-   <config evaluator="string-compare" condition="CSRFPolicy" replace="{% if use_ssl %}false{% else %}{{ csrf_config.enabled }}{% endif %}">
```
gionn, line 62: 🙄

roles/repository/tasks/main.yml
```diff
@@ -143,8 +142,17 @@
...
 
+- name: Add Hazelcast cluster config for Share
+  template:
+    owner: "{{ username }}"
+    group: "{{ group_name }}"
+    src: custom-slingshot-application-context.xml
+    dest: "{{ settings_folder }}/classpath/alfresco/web-extension/custom-slingshot-application-context.xml"
+    mode: 'u=rw,g=,o='
+  become: true
+  when: groups['repository'] | length > 1 and not cluster_keepoff
```
gionn, line 154: when there is only one node it's using an embedded hazelcast configuration?

in addition, to be truly idempotent the file needs to be absent when length <= 1

roles/nginx/templates/alfresco.conf.j2
```diff
@@ -0,0 +1,45 @@
...
+{% for repo in repo_hosts | unique %}
+{% if not repo.cluster_keepoff | default(false) %}
+        server {{ repo.local_addr | default(repo.inventory_name) }}:{{ ports_cfg.repository.http | default(8080) }} max_fails=2 fail_timeout=8s;
+{% endif %}
+{% endfor %}
+    }
+
+upstream tracker_lb {
+{% for repo in repo_hosts | unique %}
+        server {{ repo.local_addr | default(repo.inventory_name) }}:{{ ports_cfg.repository.http | default(8080) }} max_fails=2 fail_timeout=8s {% if not repo.cluster_keepoff | default(false) and repo_hosts | selectattr('cluster_keepoff', 'true') %}backup{% endif %};
```
gionn, line 19: I am not getting entirely the if here for handling backup flag:

if there are 3 nodes and one of them is marked as keepoff, this one get declared as backup

while the first condition of the if is for ensuring what?

roles/nginx/templates/alfresco.conf.j2
```diff
@@ -0,0 +1,45 @@
...
+    listen {{ item.listen | default('80') }}{% if item.listen == '443' %} ssl{% endif %};
+{% if item.server_name is defined %}
+    server_name {{ item.server_name }};
+{% endif %}
+{% if item.listen == '443' %}
+    ssl_certificate /etc/nginx/{{ item.cert_crt }};
+    ssl_certificate_key /etc/nginx/{{ item.cert_key }};
+{% endif %}
+    include {{ nginx_vhost_path }}/alfresco_proxy.include;
+}
```
gionn, line 45: the redirect from 80 to 443 is handled somewhere else?

roles/nginx/molecule/default/molecule.yml
```diff
@@ -4,7 +4,7 @@ dependency:
 driver:
   name: docker
 platforms:
-  - name: nginx
+  - name: nginx1
```
gionn, line 7: is this required?

molecule/multimachine/prepare.yml
```diff
@@ -0,0 +1,114 @@
...
+      when: molecule_yml.platforms | selectattr('name', 'in', groups.repository) | length > 1
+      block:
+        - name: get EC2 instance info
+          amazon.aws.ec2_instance_info:
+            filters:
+              "tag:Name": molecule_{{ lookup('env', 'MOLECULE_IT_PLATFORM') }}_repo_{{ lookup('env', 'BRANCH_NAME') }}_{{ lookup('env', 'BUILD_NUMBER') }}
+          register: molecule_ec2
+    
+        - name: Create Shared contentstore
+          community.aws.efs:
```
gionn, line 63: 🤯

docs/shared-contentstore.md
```diff
@@ -85,3 +85,54 @@ all:
...
+### Properly configuring your storage system
+
+Many storage technologies implement features to workaround this problem. Please see bellow some examples showing how to reliably setup shared storage for diverse clients.
+
+#### NFS
+
+This example takes advantage of the squashing mechanism NFS offers. This basically allows to map all NFS client requests to a single user (usually `nobody`) regardless of the UID/GID seen on the client.
+Configuration of a Linux NFS server explained (`/etc/exports`):
+
+````config
```
gionn, line 118:
```suggestion
```config
```

docs/shared-contentstore.md
```diff
@@ -85,3 +85,54 @@ all:
...
+  - `rw`: filesystem is available for write access (of permissions allow)
+  - `all_squash`: all client requests will be mapped to the anonymous user and group
+  - anonuid``: UID of the anonymous user (65534 is widely recognized accross many NFS clients as the nfs anonymous UID)
+  - anongid``: GID of the anonymous group (65534 is widely recognized accross many NFS clients as the nfs anonymous GID)
+
+> Of course depending on your NFS vendors options and ways of configuring them will largely differ. Please refer to your vendor documentation.
+
+Permissions on the exported filesystem would look like that on the NFS server:
+
+````bash
```
gionn, line 134:
```suggestion
```bash
```

requirements.yml
```diff
@@ -1,5 +1,7 @@
 ---
 collections:
+  - name: community.aws
+    version: 3.1.0
```
gionn, line 4: move it under `# molecule ec2 tests`

the idea was to keep requirements organized somehow

gionn (changes requested):

roles/common/defaults/main.yml
```diff
@@ -11,8 +11,17 @@ ats_home: "{{ binaries_folder }}/transform-service"
...
+  | json_query('[].
+    {
+      inventory_name: inventory_hostname , 
+      local_addr: ansible_default_ipv4.address, 
+      cluster_keepoff: cluster_keepoff
+    }') 
+  }}
+repo_host: "{{ repo_hosts[0].local_ip | default(repo_hosts[0].inventory_name) }}"
```
gionn, line 24:
```suggestion
repo_host: "{{ repo_hosts[0].local_addr | default(repo_hosts[0].inventory_name) }}"
```
?

roles/sync/molecule/default/molecule.yml
```diff
@@ -4,7 +4,7 @@ dependency:
 driver:
   name: docker
 platforms:
-  - name: sync
+  - name: instance
```
gionn, line 7: use a different instance name for each role so it's possible to converge more than one molecule test locally at the same time

roles/repository/tasks/main.yml
alxgomz: yes, kind of, absence of this file disable share clustering which is the only  purpose of hazelcast.
Regarding the second part of your comment, I think that's already what the `when:`condition` does...?

roles/repository/templates/share.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version='1.0' encoding='utf-8'?>
-<Context crossContext="true" cachingAllowed="true" cacheMaxSize="100000" docBase="{{ content_folder }}/web-server/webapps/share.war">
-  <Resources>
+<Context crossContext="true" docBase="{{ content_folder }}/web-server/webapps/share.war">
+  <Resources cacheMaxSize="20480">
```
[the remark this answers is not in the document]
alxgomz: The only reason I can think of we would need to change that is if customers has a customized Share UI with more resources to cache. Although that's not unrealistic, we're also discouraging ppl from using share in favor of ADW. But given we also have modules for share (e.g. governance) that may add resources to regular share UI, I'll just double that value which is probably a bit low.

roles/repository/vars/main.yml
```diff
@@ -15,14 +15,13 @@ share_folders:
...
 share_templates:
-  - custom-slingshot-application-context.xml
   - share-config-custom.xml
   - smartfolders-amp-actions-config.xml
 
 repository:
-  protocol: http
-  host: "{{ repo_host }}"
-  port: "{{ ports_cfg.repository.http }}"
+  protocol: http{% if use_ssl | bool %}s{% endif %}
```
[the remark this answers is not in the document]
alxgomz: ahaha I have no choice but to agree! :)

roles/sync/molecule/default/molecule.yml
alxgomz: OK I reverted but you should be able to do that jsut by cd'ing to the other folder...

roles/repository/templates/share.xml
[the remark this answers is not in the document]
gionn: keep in mind that having a knot available for fine tuning is good for uncertain things :D

roles/repository/tasks/main.yml
gionn: AFAIK `when` is a global parameter that runs or not the referenced module, so I don't think it's like having `state: absent` when `when: false`

googling doesn't help me, can you try to investigate too?

a fast solution that comes to my mind is to have another task with file module with a `when: ! $previous_task_condition` :D

roles/nginx/molecule/default/molecule.yml
gionn: we just need to decide a convention because I've started to use `role-instance` elsewhere

roles/sync/molecule/default/molecule.yml
gionn: if you do a molecule converge while you already have a molecule converge running for a different role, it tries to use the same docker container already running (I think it use the docker container name as a key) xD

roles/nginx/templates/alfresco.conf.j2
gionn: still not convinced that `not repo.cluster_keepoff | default(false)` is what you had explained, or maybe I've troubles regarding precedence of NOT operator in jijna, I try with an example:

if I have:
repo1.cluster_keepoff = false
repo2.cluster_keepoff = false
repo3.cluster_keepoff = true

I obtain:
repo1 backup
repo2 backup
repo3

what I am not getting?

alxgomz: you're getting it right. That's indeed how the template renders (and needs to). That LB is not a LB it's just a failover which prefers talking to the node that's not part of the cluster if any is present

roles/repository/tasks/main.yml
alxgomz: :-/ I won't comment on the proposed solution XD
What's the use-case you're trying to cover? The case where someone wants to move away from cluster and return to a single node? Or something else?

gionn: yeah, I think we can ignore this but was quite surprised isn't something we can easily support

roles/nginx/templates/alfresco.conf.j2
gionn: ok, the last phrase makes me understand what we are trying to achieve, I was thinking that we need to do the opposite  (flag backup only the keepoff cluster)

gionn (dismissed):

roles/nginx/molecule/default/molecule.yml
```diff
@@ -4,7 +4,7 @@ dependency:
 driver:
   name: docker
 platforms:
-  - name: nginx1
+  - name: nginx_instance
```
gionn, line 7:
```suggestion
  - name: nginx-instance
```

roles/repository/defaults/main.yml
```diff
@@ -4,3 +4,6 @@ utils_repo: []
 utils_storage:
   tmpfs: []
 cluster_keepoff: false
+
+# Tomcat resource cache size (mainly for Share)
+tocmat_resource_cache_size: 20480
```
gionn, line 9: watch out for the typo

docs/deployment-guide.md
```diff
@@ -450,6 +450,39 @@ Once ACS has initialized access the system using the following URLs with a brows
...
 
+### ACS cluster
+
+Due to load or high availability needs, you might want to deploy a cluster of several repository nodes. This can be achieved rather simply by:
+
+* Giving the playbook the location of the shared storage used for the ACS contentstore (See [Shared storage documentation](shared-contentstore.md) for details).
+* Specifying several hosts within the repository hosts group
+
+> :warning: as mention in the [Alfresco official documentation](https://docs.alfresco.com/content-services/latest/admin/cluster/#scenarioredundancycluster), "All the servers in a cluster should have static IP addresses assigned to them".
+> Not meetign this pre-requisite won't prevent the playbook from working but the cluster might will most likely stop working in case one of the server in the architecture changes IP address.
```
gionn, line 461:
```suggestion
> Not meeting this pre-requisite won't prevent the playbook from working but the cluster might will most likely stop working in case one of the server in the architecture changes IP address.
```

gionn (changes requested):
please check the currently open conversations from yesterday, the rest is fine for me

gionn (approved):

.travis.yml
```diff
@@ -27,7 +27,7 @@ env:
     - MOLECULE_NO_LOG=false
     - AWS_REGION=eu-west-1
     - MOLECULE_IT_AWS_VPC_SUBNET_ID=subnet-6bdd4223
-    - DTAS_VERSION=v1.1.1
+    - DTAS_VERSION=master
```
gionn, line 30: we need to make a release before merging if tests are passing

gionn (changes requested):

.travis.yml
alxgomz: The next planned release is 2.0 and it's not prioritized. And I'd rather not release for such a small/quick change

gionn (dismissed):

gionn (approved):

PR #270, opened 2022-02-28
[minor] Fix vagrant up

gionn:
Running `vagrant up` fails with:

```
==> default: Clearing any previously set network interfaces...
The IP address configured for the host-only network is not within the
allowed ranges. Please update the address used to be within the allowed
ranges and run the command again.

  Address: 172.100.100.100
  Ranges: 192.168.56.0/21

Valid ranges can be modified in the /etc/vbox/networks.conf file. For
more information including valid format see:

  https://www.virtualbox.org/manual/ch06.html#network_hostonly
```

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -3,7 +3,7 @@ pass     = ENV['NEXUS_PASSWORD']
 
 Vagrant.configure("2") do |config|
   config.vm.box = "centos/7"
-  config.vm.network "private_network", ip: "172.100.100.100"
+  config.vm.network "private_network", ip: "192.168.56.100"
   config.vm.provider "virtualbox" do |v|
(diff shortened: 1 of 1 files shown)


aitseitz:
I stumpled over this issue and it took me a while to start with the alfresco-ansible-deployment locally with vagrant.

+1 vote for this pullrequest

In case this pullrequest gets rejected, please update the readme.md documenation to point to
https://www.virtualbox.org/manual/UserManual.html#network_hostonly
or better provide the solution for linux based systems which is:

```
cat /etc/vbox/networks.conf   
# Network for alfresco-ansible-deployment
* 172.100.100.0/24
```

alxgomz (approved):
LGTM

PR #273, opened 2022-03-01
OPSEXP-1238 External ActiveMQ

gionn:
OPSEXP-1238

### DIFF
--- a/docs/deployment-guide.md
+++ b/docs/deployment-guide.md
@@ -216,6 +216,26 @@ The default database username (`repo_db_username` and/or `sync_db_username`) and
 
 Please refer to the [Configuring Databases](https://docs.alfresco.com/content-services/latest/config/databases/) documentation for more detailed information.
 
+### External ActiveMQ
+
+By default the playbook will deploy and configure an ActiveMQ instance that is
+suitable for testing/evaluation only (single instance and default credentials).
+
+It's strongly suggested that you provide your own ActiveMQ instance by defining
+into the inventory file one host into to `activemq` group as follows:
+
+```yaml
+    activemq:
+      hosts:
+        activemq.example.com:
+          managed_host:
+          activemq_username: alfresco
+          activemq_password: alfresco
+          activemq_port: 61617
+```
+
+The `managed_host` variable is used to prevent the playbook to manage that host.
+
 ### Custom Keystore
 
 By default the playbook deploys a default keystore to ease the installation process, however, we recommend you [generate your own keystore](https://docs.alfresco.com/content-services/latest/admin/security/#managealfkeystores) following the [instructions here](https://docs.alfresco.com/content-services/latest/admin/security/#keystore-configuration).
--- a/docs/deployment-guide.md
+++ b/docs/deployment-guide.md
@@ -222,19 +222,23 @@ By default the playbook will deploy and configure an ActiveMQ instance that is
 suitable for testing/evaluation only (single instance and default credentials).
 
 It's strongly suggested that you provide your own ActiveMQ instance by defining
-into the inventory file one host into to `activemq` group as follows:
+into the inventory file one host into to `external_activemq` group (nested
+inside the `external` group) as follows:
 
 ```yaml
-    activemq:
-      hosts:
-        activemq.example.com:
-          managed_host:
-          activemq_username: alfresco
-          activemq_password: alfresco
-          activemq_port: 61617
+all:
+  children:
+    external:
+      children:
+        external_activemq:
+          hosts:
+            whatever.mq.eu-west-1.amazonaws.com:
+              activemq_username: alfresco
+              activemq_password: alfresco
+              activemq_port: 61617
 ```
 
-The `managed_host` variable is used to prevent the playbook to manage that host.
+Every host under `external` group is skipped by the acs playbook.
 
 ### Custom Keystore
 
--- a/docs/deployment-guide.md
+++ b/docs/deployment-guide.md
@@ -218,12 +218,13 @@ Please refer to the [Configuring Databases](https://docs.alfresco.com/content-se
 
 ### External ActiveMQ
 
-By default the playbook will deploy and configure an ActiveMQ instance that is
-suitable for testing/evaluation only (single instance and default credentials).
+This playbook provides support for a single host declared inside the `activemq`
+group that will deploy and configure an ActiveMQ instance that is suitable for
+testing/evaluation only (no failover and default credentials).
 
 It's strongly suggested that you provide your own ActiveMQ instance by defining
-into the inventory file one host into to `external_activemq` group (nested
-inside the `external` group) as follows:
+into the inventory file exactly one host into to `external_activemq` group
+(nested inside the `external` group) as follows:
 
 ```yaml
 all:
@@ -238,7 +239,9 @@ all:
               activemq_port: 61617
 ```
 
-Every host under `external` group is skipped by the acs playbook.
+Every hosts under the `external` group is not directly managed by the acs
+playbook and is present in the inventory just for being referenced into the
+playbook.
 
 ### Custom Keystore
 
--- a/docs/deployment-guide.md
+++ b/docs/deployment-guide.md
@@ -237,7 +237,7 @@ all:
           activemq_port: 61617
     external:
       children:
-        external_activemq
+        external_activemq:
 ```
 
 Every hosts under the `external` group is not directly managed by the acs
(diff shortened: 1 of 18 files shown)


roles/common/defaults/main.yml
```diff
@@ -13,7 +13,7 @@ ats_home: "{{ binaries_folder }}/transform-service"
 # Using default() filter for hostvars 1st level key requires ansible 2.10+
 nginx_host: "{% if groups.nginx | length == 0 %}127.0.0.1{% else %}{{ hostvars[groups.nginx[0]].ansible_host | default('127.0.0.1') }}{% endif %}"
 repo_host: "{% if groups.repository | length == 0 %}127.0.0.1{% else %}{{ hostvars[groups.repository[0]].ansible_default_ipv4.address | default('127.0.0.1') }}{% endif %}"
-activemq_host: "{% if groups.activemq | length == 0 %}127.0.0.1{% else %}{{ hostvars[groups.activemq[0]].ansible_default_ipv4.address | default('127.0.0.1') }}{% endif %}"
+activemq_host: "{% if hostvars[groups.activemq[0]].activemq_host %}{{ hostvars[groups.activemq[0]].activemq_host }}{% else %}{{ hostvars[groups.activemq[0]].ansible_default_ipv4.address | default('127.0.0.1') }}{% endif %}"
```
gionn, line 16: probably needs an activemq_hosts

molecule/manual-activemq/hosts.yml
```diff
@@ -0,0 +1,10 @@
+all:
+  hosts:
+  children:
+    activemq:
+      hosts:
+        whatever.mq.eu-west-1.amazonaws.com:
+          managed_host: true
```
gionn, line 7:
```suggestion
          managed_host:
```

don't know if `external_host` is more explicit (managed by whom? ansible, the user or someone else?)

roles/activemq/tasks/main.yml
```diff
@@ -1,5 +1,6 @@
 ---
 # tasks file for roles/activemq
+
```
gionn, line 3:
```suggestion
```

docs/deployment-guide.md
```diff
@@ -216,6 +216,33 @@ The default database username (`repo_db_username` and/or `sync_db_username`) and
...
 Please refer to the [Configuring Databases](https://docs.alfresco.com/content-services/latest/config/databases/) documentation for more detailed information.
 
+### External ActiveMQ
+
+This playbook provides support for a single host declared inside the `activemq`
+group that will deploy and configure an ActiveMQ instance that is suitable for
+testing/evaluation only (no failover and default credentials).
+
+It's strongly suggested that you provide your own ActiveMQ instance by defining
+into the inventory file exactly one host into to `external_activemq` group
```
alxgomz, line 226:
```suggestion
in the inventory file, exactly one host as a member of the `external_activemq` group
```

docs/deployment-guide.md
```diff
@@ -216,6 +216,33 @@ The default database username (`repo_db_username` and/or `sync_db_username`) and
...
+          hosts:
+            whatever.mq.eu-west-1.amazonaws.com:
+              activemq_username: alfresco
+              activemq_password: alfresco
+              activemq_port: 61617
+```
+
+Every hosts under the `external` group is not directly managed by the acs
+playbook and is present in the inventory just for being referenced into the
+playbook.
```
alxgomz, line 244:
```suggestion
playbook and is required in the inventory just for the sake of architecture description.
```

playbooks/acs.yml
```diff
@@ -1,5 +1,5 @@
 ---
-- hosts: all
+- hosts: all:!external
```
alxgomz, line 2: as a safety measure I would actually exclude explicitly all members of `external` from each and every role

roles/activemq/molecule/default/molecule.yml
```diff
@@ -4,7 +4,7 @@ dependency:
 driver:
   name: docker
 platforms:
-  - name: activemq
+  - name: activemq-instance
```
alxgomz, line 7: the mole_it tests are using underscores (role_instance). Can we use that everywhere?

roles/activemq/molecule/default/molecule.yml
```diff
@@ -15,11 +15,19 @@ platforms:
...
+      - sfs
+      - trouter
+      - search
+      - adw
+      - activemq
 provisioner:
   name: ansible
-  ansible_args:
-    - --inventory=../../inventory_local.yml
-    - --limit=activemq
```
alxgomz, line 22: ❤️‍🔥

roles/common/defaults/main.yml
```diff
@@ -13,7 +13,12 @@ ats_home: "{{ binaries_folder }}/transform-service"
 # Using default() filter for hostvars 1st level key requires ansible 2.10+
 nginx_host: "{% if groups.nginx | length == 0 %}127.0.0.1{% else %}{{ hostvars[groups.nginx[0]].ansible_host | default('127.0.0.1') }}{% endif %}"
 repo_host: "{% if groups.repository | length == 0 %}127.0.0.1{% else %}{{ hostvars[groups.repository[0]].ansible_default_ipv4.address | default('127.0.0.1') }}{% endif %}"
-activemq_host: "{% if groups.activemq | length == 0 %}127.0.0.1{% else %}{{ hostvars[groups.activemq[0]].ansible_default_ipv4.address | default('127.0.0.1') }}{% endif %}"
+activemq_host: >-
+  {% if 'external_activemq' in groups %}
```
alxgomz, line 17:
```suggestion
  {% if groups.external_activemq %}
```
Otherwise if `external_active` is present but empty that'll break.

roles/common/defaults/main.yml
```diff
@@ -13,7 +13,12 @@ ats_home: "{{ binaries_folder }}/transform-service"
...
+  {{- groups['external_activemq'] | first -}}
+  {% else %}
+  {{- groups['activemq'] | map('extract', hostvars, ['ansible_default_ipv4','address']) | first | default('127.0.0.1') -}}
```
alxgomz, line 20: I think this 127.0.0.1 default was only usefull for previous orles' tests which where using broken molecule inventory... We should try removing them as we see them.

roles/common/defaults/main.yml
```diff
@@ -28,7 +33,12 @@ ports_cfg:
     http: "{{ ports.nginx.http | default(80) }}"
     https: "{{ ports.nginx.https | default(443) }}"
   activemq:
-    openwire: "{{ ports.activemq.openwire | default(61616) }}"
+    openwire: >-
+      {% if 'external_activemq' in groups %}
```
alxgomz, line 37: same remark as above
```suggestion
  {% if groups.external_activemq %}
```

docs/deployment-guide.md
```diff
@@ -216,6 +216,33 @@ The default database username (`repo_db_username` and/or `sync_db_username`) and
...
+all:
+  children:
+    external:
+      children:
+        external_activemq:
+          hosts:
+            whatever.mq.eu-west-1.amazonaws.com:
+              activemq_username: alfresco
+              activemq_password: alfresco
+              activemq_port: 61617
```
alxgomz, line 239: I think the below is more readable and easier to amend.
```suggestion
all:
  children:
    external_activemq:
      hosts:
        whatever.mq.eu-west-1.amazonaws.com:
          activemq_username: alfresco
          activemq_password: alfresco
          activemq_port: 61617
    external:
      children:
        external_activemq
```

alxgomz (changes requested):

roles/common/defaults/main.yml
gionn: yeah, I think I need to update all the molecule configs to include the activemq task otherwise it will complain of missing address (since no hosts in activemq group)

roles/activemq/molecule/default/molecule.yml
gionn: forgot underscore as I was mentioning before because this name is used as container/node hostname and it going to break everything

playbooks/acs.yml
gionn: don't see how it could happen if hosts are referenced by group name in the playbook

roles/common/defaults/main.yml
gionn: mine it's more defensive, when using attribute access you need to explicitly provide a `| default` or it will abort with `no such attribute` when the group is empty

playbooks/acs.yml
alxgomz: Indeed not you're right. I got lost in my reasoning here

roles/activemq/molecule/default/molecule.yml
alxgomz: true. I forgot about that. I think it's docker driver specific but better avoiding that... I wish I remembered it for my other PR 🙄

alxgomz (approved):
LGTM

PR #274, opened 2022-03-02
OPSEXP-1210 Update libreoffice version

gionn:
OPSEXP-1210


.travis.yml
```diff
@@ -240,6 +240,12 @@ jobs:
         MOLECULE_IT_SCENARIO=default
         MOLECULE_IT_CONFIG=vars-centos7.yml
 
+    - stage: Integration tests
```
alxgomz, line 243: I know you want to be as broad as possible but I think we can skip this one as we have centos7 (and RHEL8). The sole value of centos7 is that it's an rhel mirror so that seem ok to me to not test both

alxgomz (approved):
LGTM

.travis.yml
gionn: I would love to have more flexibility at pr-level, for now I suggest to postpone all the "minor" IT to master only

I understand that Centos and RHEL are very similar but in this PR there is a separated handling for centos and rhel and we wouldn't spot any regressions on one of the sides

PR #243, opened 2022-02-03
OPSEXP-1217 retry molecule test on failure

gionn:
OPSEXP-1217

PR #277, opened 2022-03-10
OPSEXP-1319 Fix random failure on async download repository task

gionn:
OPSEXP-1319

reintroduce privilege escalation for async download job checks that got removed in https://github.com/Alfresco/alfresco-ansible-deployment/pull/243


alxgomz (approved):
let's see...

PR #279, opened 2022-03-14
OPSEXP-1327 cluster rolling restarts

alxgomz:
Adds a new systemd unit in order to ensure tomcat has fully started and application is available.
This fixes concurrent restart issues and introduce a rolling restart mechanism for repo which can be useful in the future for upgrades.


roles/repository/tasks/main.yml
```diff
@@ -401,10 +401,39 @@
...
+    dest: /etc/systemd/system/alfresco-content-monitored-startup.service
+    mode: 0644
+    owner: root
+    group: root
+  notify:
+    - systemd-reload
+
+- name: Amend alfresco-content unit
+  become: true
+  lineinfile:
```
gionn, line 429: looks like a fast hack, can't we managed this in the alfresco-content.service template?

alxgomz: I didn't do it like that because the main template is part of the tomcat role. Despite its names that unit file is generic to tomcat and would work for any webapp.
This other unit on the other hand is specific for acs and as such should not be placed in the tomcat role

gionn: now I am wondering why a tomcat role setup a service called alfresco-content but that definitively not a problem of this PR

PR #284, opened 2022-03-15
OPSEXP-1330 Fix vagrant up

gionn:
OPSEXP-1330

need to abandon centos:7 since python3 is not new enough to run latest ansible

also fix OPSEXP-1331

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,3 +11,4 @@ pytest-testinfra==6.6.0
 requests==2.27.1
 yamllint==1.26.3
 dnspython==2.2.1
+Jinja2==3.0.3
--- a/roles/nginx/templates/alfresco.conf.j2
+++ b/roles/nginx/templates/alfresco.conf.j2
@@ -16,7 +16,7 @@ upstream repo_lb {
 
 upstream tracker_lb {
 {% for repo in repo_hosts | unique %}
-        server {{ repo.local_addr | default(repo.inventory_name) }}:{{ ports_cfg.repository.http | default(8080) }} max_fails=2 fail_timeout=8s {% if not repo.cluster_keepoff | default(false) and repo_hosts | selectattr('cluster_keepoff') %}backup{% endif %};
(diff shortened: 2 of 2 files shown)


Vagrantfile
```diff
@@ -1,19 +1,22 @@
-user     = ENV['NEXUS_USERNAME']
-pass     = ENV['NEXUS_PASSWORD']
-
 Vagrant.configure("2") do |config|
-  config.vm.box = "centos/7"
+  config.vm.box = "bento/ubuntu-20.04"
```
alxgomz, line 2: any reason for not using ubuntu/focal64 ?

alxgomz (commented):
Maybe we should prefer official image if possible

roles/nginx/templates/alfresco.conf.j2
```diff
@@ -16,7 +16,7 @@ upstream repo_lb {
 
 upstream tracker_lb {
 {% for repo in repo_hosts | unique %}
-        server {{ repo.local_addr | default(repo.inventory_name) }}:{{ ports_cfg.repository.http | default(8080) }} max_fails=2 fail_timeout=8s {% if not repo.cluster_keepoff | default(false) and repo_hosts | selectattr('cluster_keepoff', 'true') %}backup{% endif %}; 
+        server {{ repo.local_addr | default(repo.inventory_name) }}:{{ ports_cfg.repository.http | default(8080) }} max_fails=2 fail_timeout=8s {% if not repo.cluster_keepoff | default(false) and repo_hosts | selectattr('cluster_keepoff') %}backup{% endif %};
```
alxgomz, line 19: I've done further tests and it looks like this implicit boolean test actually evaulate every non empty string to `true`(includeing "false & "null"). This can be very missleading and we should keep the previous test from for `selectattr()`

alxgomz (changes requested):

Vagrantfile
gionn: the first one that popup on my google search =D

requirements.txt
```diff
@@ -10,5 +10,5 @@ pylint==2.12.2
 pytest-testinfra==6.6.0
 requests==2.27.1
 yamllint==1.26.3
-pyyaml==6.0
 dnspython==2.2.1
+Jinja2==3.0.3
```
gionn, line 14: Fix OPSEXP-1331

PR #289, opened 2022-03-21
Fix for jmespath conflict with boto3

gionn:
```
ERROR: Cannot install -r requirements.txt (line 7) and jmespath==1.0.0 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested jmespath==1.0.0
    boto3 1.21.19 depends on jmespath<1.0.0 and >=0.7.1
```


alxgomz (approved):
LGTM

PR #292, opened 2022-03-23
OPSEXP-1351: Implement new authentication scheme between repo & solr

alxgomz:
With that PR:

In case ACS 7.2+ is deployed, it will get the solr auth param set to `secret` and use a provided shared secret.
In case Search services 2.0.3+ is deployed, it will get the alfresco auth param set to `secret` and use a provided shared secret.
In case no shared secret is set a randomly generated one will be used.

Note: using a randomly generated password breaks the playbook idempotency. Documentation has been updated to highlight that.


playbooks/acs.yml
```diff
@@ -51,6 +51,11 @@
   roles:
     - role: '../roles/search'
   post_tasks:
+    - name: Make sure Solr is running
+      service:
+        name: alfresco-search
+        status: started
```
gionn, line 57: is this a new pattern we want to adhere with?
(ensure service are running only when running the playbook and not into the role tasks?)

maybe we want also to wait for solr really ready

gionn, line 57: is this a new pattern we want to adhere with?
(ensure service are running only when running the playbook and not into the role tasks?)

roles/common/tasks/main.yml
```diff
@@ -7,6 +7,11 @@
   include_vars: "{{ item }}"
   loop: "{{ lookup('first_found', os_fallback, errors='ignore', wantlist=True) }}"
 
+- name: Set Repo/Solr shared secret
+  set_fact:
+    reposearch_shared_secret: "{{ reposearch_shared_secret | default(lookup('password', '/dev/null')) }}"
```
gionn, line 12: can we just make it failing if a secret is not provided when required?

gionn, line 12: can we just make it failing if a secret is not provided when required?

gionn (commented):
docs update really needed 🥇

playbooks/acs.yml
alxgomz: We discussed it some time ago and I though we agreed on that approach.
Keeping that check in the role would make the service being `started` by the task (can't be a handler if we want to always run it) and then `restart` by handlers in most cases (unless there is an unexpected issue preventing the handler from being triggered).
I think that's sub-optimal and could maybe be problematic with repo service when the service first initialize the system and place clusterLock on DB.

I initially didn't think of it as a general pattern we should adhere to but your comment makes me wonder to which extend service management should be removed from the roles themselves and added to the main play?

In any case I wouldn't do that now but maybe keep it for the extensibility work

roles/common/tasks/main.yml
alxgomz: That was the other option but I decided to not follow it as we will add more and more credentials vars as we progress the "hardening" of the playbook. I'm afraid the pre-requisites to get a first play to run would be way too big.

playbooks/acs.yml
gionn: > We discussed it some time ago and I though we agreed on that approach.

repetita iuvant :D

> makes me wonder to which extend service management should be removed from the roles themselves and added to the main pla

that's probably something we need to ask to our users and how they expect after a playbook run

roles/common/tasks/secrets.yml
```diff
@@ -0,0 +1,14 @@
+---
+# tasks file for common secrets
+- name: Generate tracking/search shared secreti (none provided)
+  set_fact:
+    reposearch_shared_secret: "{{ lookup('password', '/dev/null') }}"
+  delegate_to: localhost
+  delegate_facts: true
+  run_once: true
```
gionn, line 8: is not really required anymore?

roles/search/molecule/default/molecule.yml
```diff
@@ -15,17 +15,23 @@ platforms:
...
+    groups:
+      - repository
+      - nginx
+      - search
 provisioner:
-  name: ansible
   ansible_args:
-    - --inventory=../../inventory_local.yml
+    - -e
+    - reposearch_shared_secret=dummy
```
gionn, line 25: you forgot to remove host_vars here? is set both here and there or you are testing an edge case when user set both? :D

roles/common/tasks/secrets.yml
alxgomz: removing it wouldn't break things but we would iterate against the hosts list (the hosts the repo and search roles applies to) and voerride the generated secret everytime. So that's still usefull

roles/search/molecule/default/molecule.yml
alxgomz: unfortunately I do need both the host_vars and the extra-var. The later is used for converge & idempotency to work has expected. The former is used for feature flip in testinfra tests to work (as it relies on host.ansible.get_variables which only knows about inventory vars not runtime  vars)

gionn (approved):

PR #293, opened 2022-03-23
OPSEXP-1328 Retry failing destroy during ec2 tests

gionn:
OPSEXP-1328

not sure if there is a way to just skip the cleanup phase that make sense only when running molecule locally (not on CI where the local environment is going to be discarded immediately after)

### DIFF
--- a/molecule/default/cleanup.yml
+++ b/molecule/default/cleanup.yml
@@ -8,6 +8,11 @@
     - name: Gather EC2 facts
       amazon.aws.ec2_metadata_facts:
       delegate_to: "{{ groups['nginx'] | first }}"
+      ignore_unreachable: yes
+
+    - name: End gracefully if EC2 is unavailable
+      meta: end_host
+      when: ansible_ec2_public_hostname is undefined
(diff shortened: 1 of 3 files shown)


alxgomz (changes requested):
That won't help with the issue mentioned in the ticket. No matter how many times you retry or how much time you wait. Connection to an host that's not been created won't work.
Instead we should prefer failsafe the prepare playbook for tasks depending on instance being reachable. tasks ran by localhost should fail normally

molecule/default/cleanup.yml
```diff
@@ -8,6 +8,11 @@
     - name: Gather EC2 facts
       amazon.aws.ec2_metadata_facts:
       delegate_to: "{{ groups['nginx'] | first }}"
+      ignore_unreachable: yes
+
+    - name: End gracefully if EC2 is unavailable
+      meta: end_host
```
alxgomz, line 14: very elegant

alxgomz (approved):
LGTM

PR #294, opened 2022-03-23
OPSEXP-1358 Use pre-commit from build-tools

gionn:
OPSEXP-1358


mteodori:
I still think there's no value in this shared action and I would rather not have it used

gionn:
closing since https://alfresco.atlassian.net/browse/OPSEXP-1446

PR #295, opened 2022-03-25
Minor fixes to inventories

gionn:


inventory_ha.yml
```diff
@@ -1,24 +1,27 @@
 ---
 all:
   vars:
-    ## EACH OF THE VARIABLES BELLOW CAN BE MOVED TO A GROUPS (all.children.*.vars) OR
-    ## HOSTS (all.children.group.hosts.*) SEcTIONS IN CASE VARS NEED TO DIFFER FOR EACH.
-    ## PRECEDENCE; HOSTS_VARS > GROUPS_VARS > GLOBAL_VARS.
-    # More details: https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#define-variables-in-inventory
+    # Each ssh variables below can be moved to groups (all.children.*.vars) or
```
alxgomz, line 4:
```suggestion
    # Each ssh variable below can be moved to groups (all.children.*.vars) or
```

inventory_ha.yml
```diff
@@ -1,24 +1,27 @@
...
       hosts:
         db.infra.local:
     repository:
       vars:
-        # Set the contentstore storage details as per the `mount` options
         cs_storage:
           type: nfs
           device: nas.infra.local:/exports/contentstore
+          # as per the `mount` options
+          # https://man7.org/linux/man-pages/man8/mount.8.html#FILESYSTEM-INDEPENDENT_MOUNT_OPTIONS
```
alxgomz, line 24: type and device both need to stick to the mount "dialectic" so the original location is a better fit. But maybe the comment itself wasn't clear enough to that regard and needs to be clarified

inventory_ssh.yml
```diff
@@ -1,14 +1,15 @@
 ---
 all:
   vars:
-    ## EACH OF THE VARIABLES BELLOW CAN BE MOVED TO A GROUPS (all.children.*.vars) OR
-    ## HOSTS (all.children.group.hosts.*) SEcTIONS IN CASE VARS NEED TO DIFFER FOR EACH.
-    ## PRECEDENCE; HOSTS_VARS > GROUPS_VARS > GLOBAL_VARS.
-    # More details: https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#define-variables-in-inventory
+    # Each ssh variables below can be moved to groups (all.children.*.vars) or
```
alxgomz, line 4:
```suggestion
    # Each ssh variable below can be moved to groups (all.children.*.vars) or
```

alxgomz (changes requested):

alxgomz (approved):
LGTM

PR #309, opened 2022-03-30
OPSEXP-1304 Initial support for ansible-vault

gionn:
OPSEXP-1304


alxgomz (changes requested):

docs/deployment-guide.md
```diff
@@ -213,20 +213,82 @@ If you have a valid license place your `.lic` file in the `configuration_files/l
...
+ansible-vault encrypt vars/secrets.yml
+```
+
+Now mandatory secrets are ready to use by the playbook.
+
+Please refer to the [official documentation](https://docs.ansible.com/ansible/latest/user_guide/vault.html) to learn how to interact with existing encrypted variables or files.
+
+#### Populate secrets with third-party solutions
+
+Variables defined in `vars/secrets.yml` can also be populated dynamically using third-parties lookup plugins:
```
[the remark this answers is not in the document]
gionn: we can review this together, from my understanding when the integration is via lookup plugins, the user will just declare in vars/secrets.yml something like:

```yml
activemq_password: lookup('1password', 'my/activemq/password')
```

so for the current state of the playbook, it's transparent: playbook will not try to add ansible-vault credentials because the keys are already defined in secrets.yml.

however if we expect that the user should hook a custom playbook before ours, that maybe set credentials as facts, we probably need a kill switch for out entire ansible-vault integration?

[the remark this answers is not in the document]
alxgomz: let's discuss that yes, but tbh I forgot those were lookup plugin, so that might actually be ok

docs/deployment-guide.md
```diff
@@ -213,20 +213,84 @@ If you have a valid license place your `.lic` file in the `configuration_files/l
...
+  that can automatically picked-up when running Ansible:
+
+  ```bash
+  export ANSIBLE_VAULT_PASSWORD_FILE=~/.vault_pass.txt
+  ```
+
+Otherwise you can provide the password before each playbook invocation with:
+
+```bash
+ansible-playbook --ask-vault-pass ...
```
alxgomz, line 246: Maybe worth presenting this way of enabling vaults as the main one (as it's a lot simpler) and the one using vault password file as an automation option (also because storing passwords in file won't be seen as a good default choice)

docs/deployment-guide.md
```diff
@@ -213,20 +213,84 @@ If you have a valid license place your `.lic` file in the `configuration_files/l
...
+#### Populate secrets with third-party solutions
 
-This secret should be placed either in the inventory file under the `all` group scope, or passed as an extra variable (it needs to be available to the localhost's hostvars array of variables)
+Variables defined in `vars/secrets.yml` can also reference remote values using
+third-parties lookup plugins:
 
-:warning: Should you forget to provide that shared secret, the playbook will generate a random one. While that may sound convenient keep in mind that doing so will break the idempotency of the playbook and the shared secret will be updated every time you run the playbook.
+* [HashiCorp Vault](https://docs.ansible.com/ansible/latest/collections/community/hashi_vault/hashi_vault_lookup.html)
+* [AWS Secrets](https://docs.ansible.com/ansible/latest/collections/amazon/aws/aws_secret_lookup.html)
+* [1Password](https://docs.ansible.com/ansible/latest/collections/community/general/onepassword_lookup.html)
```
alxgomz, line 293: I think Valerio mentionned this one wsa used by our custoemrs

```suggestion
* [1Password](https://docs.ansible.com/ansible/latest/collections/community/general/onepassword_lookup.html)
* [CyberArk](https://docs.ansible.com/ansible/latest/collections/community/general/cyberarkpassword_lookup.html)
```

playbooks/secrets.yml
```diff
@@ -0,0 +1,114 @@
...
+    secrets_file: "{{ base_folder }}/vars/secrets.yml"
+  tasks:
+    - name: Ensure no secrets are still set has group_vars/hostvars
+      assert:
+        that:
+          - hostvars[item].repo_db_password is not defined
+          - hostvars[item].sync_db_password is not defined
+          - hostvars[item].reposearch_shared_secret is not defined
+        quiet: true
+      loop: "{{ groups['database'] | default([]) + groups['repository'] | default([]) }}"
```
alxgomz, line 15: I think this really needs to be a tasks assigned to all hosts but external. 
* avoid maintaining the loop items
* provides wider control on vars leaking secrets on unexpected hosts

playbooks/secrets.yml
```diff
@@ -0,0 +1,114 @@
...
+          when: not lookup('vars', item, default=False)
+          loop:
+            - repo_db_password
+            - sync_db_password
+            - reposearch_shared_secret
+            - activemq_password
+
+        - name: Initialize encrypted file if requested via vault_file_initialize
+          ansible.builtin.command: ansible-vault encrypt {{ secrets_file }}
+          when: vault_file_initialize is defined
```
alxgomz, line 52: As we are in a block statement, it could be worth making this task part of an `always` statement so we are sure the plaintext file is never left unencrypted even in case of failure of previous task (https://docs.ansible.com/ansible/latest/user_guide/playbooks_blocks.html#handling-errors-with-blocks)

playbooks/secrets.yml
```diff
@@ -0,0 +1,114 @@
...
+        quiet: true
+
+    - name: Ensure activemq_password is set when required
+      assert:
+        that: activemq_password is defined and activemq_password | length > 0
+        msg: "activemq_password must have been already set at this point"
+        quiet: true
+      when: groups.external_activemq | default(False)
+
+- hosts: all:!external
```
alxgomz, line 74: why assigning this task to all but external and then skip all but expected targets?
Let's just run the tasks where we really need it

playbooks/secrets.yml
```diff
@@ -0,0 +1,114 @@
...
+  gather_facts: false
+  tasks:
+    - name: Set tracking/search shared secret
+      set_fact:
+        reposearch_shared_secret: "{{ hostvars.localhost.reposearch_shared_secret }}"
+      when: |
+        inventory_hostname in groups.repository | default([]) +  groups.search | default([])
+        and hostvars.localhost.reposearch_shared_secret is defined
```
alxgomz, line 82: Do we really need it that now we have an `assert` task?

playbooks/secrets.yml
```diff
@@ -0,0 +1,114 @@
...
+      when: |
+        inventory_hostname in groups.repository | default([]) + groups.database | default([])
+        and hostvars.localhost.repo_db_password is defined
+
+    - name: Set sync_db_password secret
+      set_fact:
+        sync_db_password: "{{ hostvars.localhost.sync_db_password }}"
+      when: |
+        inventory_hostname in groups.syncservice | default([]) + groups.database | default([])
+        and hostvars.localhost.sync_db_password is defined
```
alxgomz, line 96: same as above comments

playbooks/secrets.yml
```diff
@@ -0,0 +1,114 @@
...
+      set_fact:
+        activemq_username: "{{ groups.external_activemq | map('extract', hostvars, ['activemq_username']) | first }}"
+        activemq_password: "{{ hostvars.localhost.activemq_password }}"
+      when: |
+        groups.external_activemq | default(False) and
+        hostvars.localhost.activemq_password is defined and
+        inventory_hostname in
+          groups.syncservice | default([]) +
+          groups.repository | default([]) +
+          groups.transformers | default([])
```
alxgomz, line 108: same as above comments

alxgomz (changes requested):

playbooks/secrets.yml
```diff
@@ -0,0 +1,85 @@
...
+    - name: Ensure no secrets are still provided via group_vars and/or hostvars
+      assert:
+        that:
+          - repo_db_password is not defined
+          - sync_db_password is not defined
+          - reposearch_shared_secret is not defined
+        msg:
+          "The current playbook version expect secrets to be provided in vars/secrets.yml.
+          If you are upgrading from a previous playbook version, please read to docs/playbook-upgrade.md"
+        quiet: true
```
alxgomz, line 14: main playbook should stop if any host has such an env.
Here I think if one var is present on a single host only that host will be taken out of the play but play will go on.

playbooks/secrets.yml
```diff
@@ -0,0 +1,85 @@
...
+          If you are using encrypted files you need to add them manually before the play."
+        quiet: true
+
+    - name: Ensure activemq_password is set when required
+      assert:
+        that: activemq_password is defined and activemq_password | length > 0
+        msg: "activemq_password must have been already set at this point.
+          If you are using encrypted files you need to add manually the secret before the play."
+        quiet: true
+      when: groups.external_activemq | default(False)
```
alxgomz, line 49: We should have that even if not using external activemq

alxgomz (changes requested):

playbooks/secrets.yml
```diff
@@ -0,0 +1,85 @@
...
+          If you are upgrading from a previous playbook version, please read to docs/playbook-upgrade.md"
+        quiet: true
+
+- hosts: localhost
+  gather_facts: false
+  vars:
+    base_folder: "{{ playbook_dir }}/.."
+    secrets_file: "{{ base_folder }}/vars/secrets.yml"
+  tasks:
+    - name: Ensure empty secrets.yml is present
```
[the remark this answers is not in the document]
alxgomz: @gionn @vprovaggi If I understand correctly, this approach prevents completely kick starting the playbook without first generating the secret file manually.
That worries me a bit tbh.
Can we discuss that ?

playbooks/secrets.yml
gionn: is currently not possible to configure credentials with our activemq role (see the [next task](https://github.com/Alfresco/alfresco-ansible-deployment/pull/309/files#diff-90d24c1b1d1fdebb80688e0327d87085edba291742a1e7775f0d673ecd77d00fR81)) so there is no reason to set credential if not using external activemq

playbooks/secrets.yml
gionn: reproduced creating a `group_vars/database.yml` with a credential in it, `any_errors_fatal` helps

docs/deployment-guide.md
```diff
@@ -213,20 +213,123 @@ If you have a valid license place your `.lic` file in the `configuration_files/l
...
+### Secrets management
 
-As of ACS 7.2 and/or Search services 2.0.3, the repository <--> solr communication requires to be authenticated. The playbook will set up that authentication scheme using the new `secret` method.
-This methods needs to be passed a shared secret. In order to do so use the variable below:
+This playbook expects that security-relevant variables are configured within the
+`vars/secrets.yml` file.
 
-```yaml
-reposearch_shared_secret: dummy
+It is strongly recommended to enable [Ansible
```
vprovaggi, line 221: Not clear if Ansible vault is a recommendation or a prerequisite.

In order to avoid keeping secrets in plain text on the control node It is strongly recommended to enable [Ansible
 Vault](https://docs.ansible.com/ansible/latest/user_guide/vault.html) encryption. Though [third-party plugins](#third-party-lookup-plugins) can be used, the rest of the guide provide example for Ansible Vault.
 or use .

gionn: credentials in secrets.yml are mandatory, while vault is not

playbooks/secrets.yml
```diff
@@ -14,6 +14,10 @@
           If you are upgrading from a previous playbook version, please read to docs/playbook-upgrade.md"
         quiet: true
 
+- name: Autogenerate secrets if this is a testing environment
+  ansible.builtin.import_playbook: secrets-init.yml
+  when: test_environment is defined and test_environment|bool
```
alxgomz, line 19: Please a choose a variable name that really sas what the param does. `test_environment`sounds like an option that's not wise to use.
I had in mind stuff like `autogen_secrets` or similar.

playbooks/secrets.yml
alxgomz: OK I created OPSEXP-1459 and linked it to that same epic

alxgomz (changes requested):

playbooks/secrets.yml
gionn: I think about it twice and it's better if it's self-explanatory, for better understanding of everyone

alxgomz: Do you mean `test_environment` is self-explainatory?
If so, I have to disagree on that. allowing using autogenerated secrets is not linked only to test scenarios

gionn: sorry, the subject was your proposal, I am switching to `autogen_unsecure_secrets` to enforce that is not a good practice to enable it

alxgomz (approved):
LGTM

PR #311, opened 2022-04-01
OPSEXP-1373 Use different users for different databases

gionn:
OPSEXP-1373

### DIFF
--- a/.github/workflows/community.yml
+++ b/.github/workflows/community.yml
@@ -1,14 +1,14 @@
 name: "community"
 on:
   pull_request:
-      paths-ignore:
-      - 'docs/**'
-      - '*.md'
+    paths-ignore:
+      - "docs/**"
+      - "*.md"
   push:
     branches: [master]
     paths-ignore:
-      - 'docs/**'
-      - '*.md'
+      - "docs/**"
+      - "*.md"
 jobs:
   docker:
     runs-on: ubuntu-latest
@@ -16,17 +16,17 @@ jobs:
       fail-fast: true
       matrix:
         molecule_distro:
-          - { "image": "centos:7" }
-          - { "image": "ubuntu:20.04" }
+          - image: centos:7
+          - image: ubuntu:20.04
         role:
-          - { "name": "activemq"}
-          - { "name": "common" }
-          - { "name": "java"}
-          - { "name": "nginx"}
-          - { "name": "postgres"}
-          - { "name": "search"}
-          - { "name": "tomcat"}
-          - { "name": "transformers"}
+          - name: activemq
+          - name: common
+          - name: java
+          - name: nginx
+          - name: postgres
+          - name: search
+          - name: tomcat
+          - name: transformers
     env:
       PY_COLORS: 1
     steps:
--- a/.github/workflows/community.yml
+++ b/.github/workflows/community.yml
@@ -23,6 +23,7 @@ jobs:
           - { "name": "common" }
           - { "name": "java"}
           - { "name": "nginx"}
+          - { "name": "postgres"}
           - { "name": "search"}
           - { "name": "tomcat"}
           - { "name": "transformers"}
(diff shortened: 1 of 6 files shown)


alxgomz (changes requested):
Can you please add a doc section for people upgrading the playbook so they know they might want to revoke privileges from previous user alfresco on the alfresco-sync db?

gionn:
> Can you please add a doc section for people upgrading the playbook so they know they might want to revoke privileges from previous user alfresco on the alfresco-sync db?

I've handled that in the task itself

roles/postgres/tasks/main.yml
```diff
@@ -54,39 +54,47 @@
...
         password: "{{ repo_db_password }}"
         priv: "ALL"
         expires: infinity
         role_attr_flags: SUPERUSER
-      no_log: true
       tags:
         - molecule-idempotence-notest
     
-    - name: "Connect to {{ sync_db_name }} database and set user's password with no expire date"
+    - name: "Give {{ sync_db_username }} user to access {{ sync_db_name }} database with password"
```
mteodori, line 81: user to access => user access

roles/postgres/tasks/main.yml
```diff
@@ -54,39 +54,47 @@
...
         name: "{{ sync_db_username }}"
         password: "{{ sync_db_password }}"
         priv: "ALL"
         expires: infinity
         role_attr_flags: SUPERUSER
-      no_log: true
+      tags:
+        - molecule-idempotence-notest
+
+    - name: "Revoke {{ repo_db_username }} user to access {{ sync_db_name }} database due to OPSEXP-1373"
```
mteodori, line 92: user to access => user access
let's not reference internal issues in code as external users don't have access and can't get its meaning

mteodori (changes requested):

.github/workflows/community.yml
```diff
@@ -23,6 +23,7 @@ jobs:
           - { "name": "common" }
           - { "name": "java"}
           - { "name": "nginx"}
+          - { "name": "postgres"}
           - { "name": "search"}
           - { "name": "tomcat"}
           - { "name": "transformers"}
```
alxgomz, line 29: Why not using simple yaml notation?

```yaml
role:
  - name: activemq
  - name: common
```

roles/postgres/tasks/main.yml
alxgomz: or actually, user to access => user access to :)

roles/postgres/tasks/main.yml
```diff
@@ -54,39 +54,47 @@
...
-    - name: "Connect to {{ repo_db_name }} database and set user's password with no expire date"
+    - name: "Give {{ repo_db_username }} user to access {{ repo_db_name }} database with password"
       postgresql_user:
         db: "{{ repo_db_name }}"
         name: "{{ repo_db_username }}"
         password: "{{ repo_db_password }}"
         priv: "ALL"
         expires: infinity
         role_attr_flags: SUPERUSER
-      no_log: true
```
alxgomz, line 78: How do we ensure running ansible in debug doesn't expose sensitive data then?

roles/postgres/molecule/default/converge.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: Converge
+  hosts: all
+  tasks:
+    - name: "Include roles/postgres"
+      include_role:
+        name: "postgres"
```
alxgomz, line 7: nice! far better than --limit provisioner! 🏅

alxgomz (commented):
Mostly cosmetic stuffs

roles/postgres/tasks/main.yml
gionn: since it's an upstream module I hope they are taking care of it - looks like so https://github.com/ansible-collections/community.postgresql/blob/218641f24473574c3d59cad2b5bb5b9ad886982f/plugins/modules/postgresql_user.py#L905

alxgomz (approved):
LGTM

PR #312, opened 2022-04-04
Test vagrant with macos workflow

gionn:
https://stackoverflow.com/a/66261102/547195


gionn:
Technically it works but macos hosted runner costs 10x linux runners minutes, so it's probably not worth it.

PR #313, opened 2022-04-04
OPSEXP-1375 Fix propagation of activemq credentials when external activemq is set

gionn:
* enforce mutual exclusion of external activemq and activemq in a dedicated task
* add ec2 integration test
* fix credentials and ssl support in repository, sync, transformers

OPSEXP-1375


roles/common/tasks/prerequisites.yml
```diff
@@ -0,0 +1,9 @@
+---
+- name: Fail if external and local activemq are both in use
+  fail:
+    msg: Detected both external_activemq and activemq groups in use
+  when:
+    - groups.external_activemq is not empty
+    - groups.activemq is not empty
+  tags:
+    - molecule-notest
```
alxgomz, line 9: I have the feeling this has been done just to allow using activemq host in the molecule inventory for tests, I think that's do-able and preferable using molecule prepare playbook.

roles/common/tasks/secrets.yml
```diff
@@ -12,3 +12,9 @@
   set_fact:
     reposearch_shared_secret: "{{ hostvars.localhost.reposearch_shared_secret }}"
   when: inventory_hostname in groups.repository + groups.search
+
+- name: Override external activemq credentials if present
```
alxgomz, line 16:
```suggestion
- name: Set external activemq credentials if present
```

roles/sync/templates/config.yml.j2
```diff
@@ -3,8 +3,8 @@ version: {{ sync.version }}
 
 # ActiveMQ config
 messaging:
-    username: {{ activemq_user | default('admin') }}
-    password: {{ activemq_password | default('admin') }}
+    username: {{ activemq_username }}
```
alxgomz, line 6: template will break if no external host is set and activemq_* param are present.

alxgomz (changes requested):

group_vars/all.yml
```diff
@@ -117,6 +117,9 @@ sync_db_name: "alfresco-sync"
 sync_db_username: "alfresco"
 sync_db_password: "alfresco" # pragma: allowlist secret
 
+activemq_username: "admin"
+activemq_password: "admin" # pragma: allowlist secret
```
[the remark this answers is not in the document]
alxgomz: If that's really a big problem for OPSEXP-1304 we can explicitly exclude the pass from the ticket and create another one dedicated to it?

roles/common/tasks/prerequisites.yml
gionn: yes, it's a shameless workaround to relax that check just for molecule otherwise I can't provision activemq and use external activemq feature.

not sure it's really easy to do everything in the prepare playbook, I need to craft an inventory, startup an ec2 instance, run the playbook...?

molecule/external-activemq/cleanup.yml
```diff
@@ -0,0 +1,23 @@
...
+
+    - name: Destroy mq broker
+      ansible.builtin.shell: |
+        aws mq delete-broker --broker-id {{ describe_broker_output.stdout }}
+      when: describe_broker_output.stdout | length > 0
+
+    - name: Delete EC2 security group for mq broker
+      amazon.aws.ec2_group:
+        name: molecule-{{ lookup('env', 'BUILD_NUMBER') }}
+        state: absent
```
gionn, line 23: need to wait for broker termination or it won't work

molecule/external-activemq/prepare.yml
```diff
@@ -0,0 +1,55 @@
...
+  tasks:
+    - name: Create EC2 security group for mq broker
+      amazon.aws.ec2_group:
+        name: "{{ broker_name }}"
+        description: Molecule external activemq EC2 test
+        rules:
+          - proto: tcp
+            from_port: 61617
+            to_port: 61617
+            cidr_ip: "{{ ansible_ec2_local_ipv4 }}/32"
```
alxgomz, line 19: In that case nginx and repo are the same but that's generally not true. At least comment the assumption or state this won't work with multimachines. Otherwise using the full subnet is not a problem imho

molecule/external-activemq/prepare.yml
```diff
@@ -0,0 +1,55 @@
...
+      retries: 20
+      delay: 60
+
+    - name: Grab activemq hostname
+      ansible.builtin.shell: |
+        aws mq describe-broker --broker-id {{ broker_name }} | \
+          jq -r '.BrokerInstances[0].Endpoints[0]' | \
+          awk -F/ '{print $3}' | \
+          sed 's/:61617//'
+      register: activemq_hostname
```
alxgomz, line 49: lol I'm surprised you duplicated this task. Why not using a simple avoiding filtering previous task with jq andparsing registred json?

roles/activemq/tasks/main.yml
```diff
@@ -94,8 +94,8 @@
     - name: Add activemq wrapper script
       become: true
       template:
-        src: "activemq.sh"
-        dest: "{{ binaries_folder }}"
+        src: activemq.sh.j2
```
alxgomz, line 97: 🙏

roles/common/defaults/main.yml
```diff
@@ -10,9 +10,9 @@ ats_home: "{{ binaries_folder }}/transform-service"
 
 # Due to https://github.com/ansible/ansible/issues/64957
 # Using default() filter for hostvars 1st level key requires ansible 2.10+
-nginx_host: "{{ groups['nginx'] | map('extract', hostvars, ['ansible_default_ipv4','address']) | first | default('127.0.0.1') }}"
+nginx_host: "{{ groups['nginx'] | map('extract', hostvars, ['ansible_default_ipv4','address']) | first | default('127.0.0.1') }}"
```
alxgomz, line 13: those are my bad! I really ned to understand why that happen to me!

roles/common/defaults/main.yml
```diff
@@ -43,7 +47,7 @@ ports_cfg:
   activemq:
     openwire: >-
       {% if groups.external_activemq | default(False) %}
-      {{- groups.external_activemq[0].activemq_port | default(61616) -}}
```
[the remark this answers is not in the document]
alxgomz: gosh missed it too

roles/common/tasks/secrets.yml
```diff
@@ -16,3 +16,15 @@
...
+- name: Set external activemq credentials if present
+  set_fact:
+    activemq_username: "{{ groups.external_activemq | map('extract', hostvars, ['activemq_username']) | first }}"
+    activemq_password: "{{ groups.external_activemq | map('extract', hostvars, ['activemq_password']) | first }}"
+  when: groups.external_activemq | default(False)
+
+- name: Fallback to default insecure credentials without external activemq
+  set_fact:
+    activemq_username: admin
+    activemq_password: admin
```
alxgomz, line 29: Let's do it like that for now, but that's the reason why I think we need a proper separate secret playbook we just import rather than something in the common role. That way we can make default password generation happen outside of the main playbook leaving room for ppl to amend generated password if they want to and don't need to rely on default stuff

alxgomz (commented):
mostly comments

molecule/external-activemq/prepare.yml
gionn: correct, but since this scenario cover just a single instance it should be good to go

or maybe we can cover the external activemq in the multi-node scenario instead of using this dedicated scenario? 🤔

roles/common/defaults/main.yml
gionn: non-printable characters, who knows xD

molecule/external-activemq/prepare.yml
alxgomz: that would actually make sense to me. Up to you if you prefer merging as is it's fine too.

alxgomz (approved):
I leave it to you to decide wether you want to move the test to the multi-machine scenario

gionn:
> I leave it to you to decide wether you want to move the test to the multi-machine scenario

I want to do that but in a new task xD

PR #320, opened 2022-04-14
OPSEXP-1395: controlled usage of become  within repo role

alxgomz:
make sure become is not used at high level unwilling propagating to tasks without really realizing it.


molecule/default/molecule.yml
```diff
@@ -27,6 +27,9 @@ platforms:
 
 provisioner:
   name: ansible
+  config_options:
+    defaults:
+      pipelining: True
```
gionn, line 32: this is the workaround provided in https://docs.ansible.com/ansible/latest/user_guide/become.html#risks-of-becoming-an-unprivileged-user?

if we want this to make everything work we should document it for the user too

env var could be of help in multiple use case
https://docs.ansible.com/ansible/latest/reference_appendices/config.html#ansible-pipelining

vagrant included

roles/activemq/tasks/main.yml
```diff
@@ -1,138 +1,114 @@
...
+      blockinfile:
+        path: "{{ config_folder }}/setenv.sh"
+        marker: "# {mark} ACTIVEMQ ENV VARS"
+        owner: "{{ username }}"
+        group: "{{ group_name }}"
+        block: |
+          export ACTIVEMQ_HOME={{ activemq_home }}
+          export ACTIVEMQ_HOST={{ activemq_host }}
+      tags:
+        - molecule-idempotence-notest
```
gionn, line 94: blockinfile isn't idempotent?

roles/common/tasks/secrets.yml
```diff
@@ -6,9 +6,13 @@
...
+    hostvars.localhost.reposearch_shared_secret is undefined
+    and groups.search | default([])
 
 - name: Set tracking/search shared secret
   set_fact:
     reposearch_shared_secret: "{{ hostvars.localhost.reposearch_shared_secret }}"
-  when: inventory_hostname in groups.repository + groups.search
+  when: |
+    inventory_hostname in groups.repository | default([]) +  groups.search | default([])
+    and hostvars.localhost.reposearch_shared_secret is defined
```
gionn, line 18: this was bothering me too 🥇

roles/java/tasks/main.yml
```diff
@@ -1,55 +1,32 @@
...
-  when: not setenv_exists.stat.exists
+    - name: Add setenv bash file
+      template:
+        src: setenv.sh
+        dest: "{{ config_folder }}/setenv.sh"
+        owner: "{{ username }}"
+        group: "{{ group_name }}"
+        mode: "0755"
+      tags:
+        - molecule-idempotence-notest
```
gionn, line 32: if variables used in the template doesn't changes between run it should be idempotent 🤔

gionn (changes requested):

molecule/default/molecule.yml
alxgomz: I saw it mostly as a speedup measure. I actually though it was already in place, but due to molecule using it's own ansible.cfg it slipped off.
Having said that the postgres role requires pipelining for `postgres_user` tasks. No specific action is required by the user as it's explicit in the tasks but some sudo config may prohibit that.
Given postgres role is only provided for convenience and we offer the ability to use an external db for prod env. I think it's fair to document the requirement for the postgres  host.

roles/activemq/tasks/main.yml
alxgomz: this isolated task and the  module itself are indeed idempotent. But the file is also handled in the java role using a `template` task. I think that's was created the idempotency issue.
However given I later on made this other taks skip idempotency checks I guess it's OK to revert this one to normal.
Generally speaking, we could agree that in such scenarios we should:
1. tag the root tasks only 
2. migrate to a model using a set of `template` + `assemble` tasks

I'm open to idea.
I think that basically links to OPSEXP-1280 imho

roles/java/tasks/main.yml
alxgomz: see my previous comment. It's all about handling the same file in multiple roles

gionn (approved):

PR #321, opened 2022-04-15
OPSEXP-1401 controlled become usage within the transformer roles

alxgomz:
make sure become is not used at high level unwilling propagating to tasks without really realizing it.


gionn (approved):
* looks like this PR is based on the previous PR we already merged, in this case please rebase the branch before review otherwise I am reviewing same things two times
* since it's a big PR and most changes are refactoring (or a sort of), try to outline with a comment in the PR the most interesting parts for the review that are not just refactoring - if any

PR #331, opened 2022-04-25
OPSEXP-1424: fix setenv.sh being wiped by subsequent roles' plays

### LINKED ISSUE #330: Missing environment variables TOMCAT_HOME
#### Bug description

After resolving issues in https://github.com/Alfresco/alfresco-ansible-deployment/issues/328 and not handling errors reported in https://github.com/Alfresco/alfresco-ansible-deployment/issues/329, I can see that Alfresco cannot startup because of missing environment variables.

If I start Alfresco repository by `sudo systemctl start alfresco-content.service`, I get error message visible using `journalctl -xe`: 
```shell
Apr 23 11:46:40 tomcat.sh[31057]: /bin/bash: /bin/catalina.sh: No such file or directory
Apr 23 11:46:40 systemd[1]: alfresco-content.service: Control process exited, code=exited status=127
Apr 23 11:46:40 systemd[1]: alfresco-content.service: Failed with result 'exit-code'.
```

The same error message can be seen if I run startup command directly by alfresco user:
```shell
$ /opt/alfresco/tomcat.sh start
/bin/bash: /bin/catalina.sh: No such file or directory
```

The problem is caused by missing environment variables. If I run part of tomcat.sh manually, I can see that value for "CATALINA_HOME" is empty:
```shell
$ . /etc/opt/alfresco/setenv.sh
$ export CATALINA_HOME=${TOMCAT_HOME}
$ env | grep CATA
CATALINA_HOME=
```

The problem here is caused already by missing "TOMCAT_HOME"

#### Target OS

RHEL 8.3

#### Ansible error

The error is not visible during Ansible deploy but after that once Alfresco repository is started.

#### Ansible context

Same as https://github.com/Alfresco/alfresco-ansible-deployment/issues/328

alxgomz:
Adds tests to existing scenarios

fixes #330
Ref. OPSEXP-1424

### DIFF
--- a/roles/java/tasks/main.yml
+++ b/roles/java/tasks/main.yml
@@ -22,11 +22,15 @@
           - --xform=s|[^/]*|{{ java_home | basename }}|
 
     - name: Add setenv bash file
-      template:
-        src: setenv.sh
-        dest: "{{ config_folder }}/setenv.sh"
+      blockinfile:
+        path: "{{ config_folder }}/setenv.sh"
+        marker: "# {mark} JAVA ENV VARS"
+        create: yes
+        block: |
+          #!/bin/sh
+          #
+          export JAVA_HOME="{{ java_home }}"
+          export PATH="${JAVA_HOME}/bin:${PATH}"
         owner: "{{ username }}"
         group: "{{ group_name }}"
         mode: "0755"
-      tags:
-        - molecule-idempotence-notest
--- a/molecule/default/side_effects.yml
+++ b/molecule/default/side_effects.yml
@@ -1,0 +1,3 @@
+---
+- name: Restart all services in order
+  import_playbook: ../../playbooks/platform-restart.yml
--- a/playbooks/platform-restart.yml
+++ b/playbooks/platform-restart.yml
@@ -1,0 +1,186 @@
+---
+- hosts: adw
+  gather_facts: no
+  become: true
+  tasks:
+    - name: Stop Alfresco Digital Workspace
+      service:
+        name: nginx
+        state: stopped
+      tags:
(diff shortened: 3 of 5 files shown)


roles/java/tasks/main.yml
```diff
@@ -22,11 +22,15 @@
...
-      template:
-        src: setenv.sh
-        dest: "{{ config_folder }}/setenv.sh"
+      blockinfile:
+        path: "{{ config_folder }}/setenv.sh"
+        marker: "# {mark} JAVA ENV VARS"
+        create: yes
+        block: |
+          #!/bin/sh
+          #
```
gionn, line 31: I've the feeling it's more robust if we define a separate task which deploy a static file with just the shebang instead of relying that this will be forever the first task creating that file (`create: yes` will mask the file created before this task)

alxgomz: In that case the issue remains. The java role being a dependency of other role using the setenv.sh, the playbook would endup wiping the setenv.sh file for every host with at least 2 roles depending on java.
The alternative I really would like to implement at some point is to completely stop relying on such a hacky setenv.sh script and instead make sure to populate the systemd unit files with the right ENV vars (either in the unit file itself or using an env file).

gionn: yeah you are right, probably blockinfile should be used only for files deployed externally from ansible and we should instead have a template for that file

gionn (approved):

roles/java/tasks/main.yml
alxgomz: created OPSEXP-1430 to track that

PR #336, opened 2022-05-02
OPSEXP-1376 Fix postgres users too broad privileges

gionn:
OPSEXP-1376

* application users are not superusers anymore
* access to default schema is revoked https://dba.stackexchange.com/a/35317/5179

### DIFF
--- a/docs/deployment-guide.md
+++ b/docs/deployment-guide.md
@@ -275,6 +275,10 @@ By default the playbook will deploy and configure a Postgres server for you. Tha
 
 > This server also requires to NOT have a sudo configuration with `requirestty` set.
 
+Please be aware that a change in Postgres configuration (e.g. pg_hba) can
+trigger a service restart and application downtime. For this reason you may want
+to run the playbook only during a scheduled maintenance window.
+
 If you'd prefer to use an external database server you can override the `repo_db_url` variable.
 
 An example custom database url is shown below:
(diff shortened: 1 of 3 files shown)


roles/postgres/tasks/main.yml
```diff
@@ -62,20 +62,35 @@
...
-      postgresql_db:
+      community.postgresql.postgresql_db:
         name: "{{ item }}"
       loop:
         - "{{ repo_db_name }}"
         - "{{ sync_db_name }}"
 
-    - name: Set users privileges on databases
-      postgresql_user:
+    # https://dba.stackexchange.com/a/35317/5179
```
alxgomz, line 71: I would rather put that in the JIRA's & PR comments rather than the code itself. So people can always track back to it using git blame or github PR.
Having it in the code can make sense if it allows pointing something that needs to later be changed but here that's just how things have to work anyway

roles/postgres/tasks/main.yml
```diff
@@ -62,20 +62,35 @@
...
+    - name: Revoke public access to databases
+      community.postgresql.postgresql_privs:
+        db: "{{ item }}"
+        privs: ALL
```
alxgomz, line 75: This (and the fact we restart the services during play) enforces someone have to run the playbook during maintenance schedules only. That's an acceptable constraint but we should document it now maybe.

roles/postgres/tasks/main.yml
```diff
@@ -62,20 +62,35 @@
...
+        objs: public
+        role: public
+        state: absent
+      loop:
+        - "{{ repo_db_name }}"
+        - "{{ sync_db_name }}"
+      tags:
+        - molecule-idempotence-notest
+
+    - name: Create deprivileged users
```
alxgomz, line 86:
```suggestion
    - name: Create unprivileged users
```

alxgomz (changes requested):

docs/deployment-guide.md
```diff
@@ -275,6 +275,10 @@ By default the playbook will deploy and configure a Postgres server for you. Tha
 
 > This server also requires to NOT have a sudo configuration with `requirestty` set.
 
+Please be aware that a change in Postgres configuration (e.g. pg_hba) can
```
alxgomz, line 278: Sorry I think my previous comment was unclear. Actually the reset of permissions  is just another thing that makes the whole playbook not usable outside of maintenance slots. Hence this comment applies are the playbook level not only for db tasks

alxgomz (changes requested):

docs/deployment-guide.md
gionn: wait, I was thinking that you are referring to restart problems in general and not specific to this pr: here I am changing permissions on the fly and no service will be restarted.

in any case I think it's fine to add something to upper docs

PR #337, opened 2022-05-02
OPSEXP-1438 Bump openjdk to 11.0.15

gionn:
OPSEXP-1438


group_vars/all.yml
```diff
@@ -78,12 +78,14 @@ amp_downloads:
 # and change this flag to true. See the deployment guide for more details.
 use_custom_keystores: false
 
+# common dependencies should be updated also in *-extra-vars.yml as well thanks
+# to default ansible merging behaviour
 dependencies_version:
```
gionn, line 83: I think in the end it's better to avoid nested objects for variables when you don't really need to manage all of them as they were one - like here

alxgomz: idk... I don't like a lot having plenty fo flat vars. For now I actually even consider we should deal with all dependencies data (name, url, checksums, version and so on...) in one single data structure... That may make our life less easy but gives user some structure and clarity when it comes to configuring the playbook... Anyway that's not our topic here but I just wanted to comment

alxgomz (approved):
LGTM

PR #338, opened 2022-05-02
Switch to openjdk jre

gionn:


.prospector.yml
```diff
@@ -13,7 +13,6 @@ pylint:
     - R1705
     - R1723
     - W0511
-    - W0613
```
gionn, line 16: https://pycodequ.al/docs/pylint-messages/w0613-unused-argument.html

mteodori (approved):

PR #341, opened 2022-05-04
OPSEXP-1445 Merge external activemq into clustering integration test

gionn:
OPSEXP-1445


molecule/default/create.yml
```diff
@@ -10,7 +10,7 @@
     - community.crypto
   vars:
     # Run config handling
-    default_run_id: "{{ lookup('password', '/dev/null chars=ascii_lowercase length=5') }}"
+    default_run_id: "{{ lookup('env', 'BUILD_NUMBER') }}"
```
gionn, line 13: FIXME needs to be different for each job in the ec2 matrix otherwise it will fail

molecule/default/create.yml
```diff
@@ -209,6 +209,7 @@
         index_var: index
         label: "{{ item.name }}"
       when: item.security_groups | length == 0
+      register: ec2_instances_group
```
gionn, line 212:
```suggestion
```

Mieszko96 (approved):
LGTM

PR #342, opened 2022-05-05
OPSEXP-1305 Handle custom keystore password

gionn:
OPSEXP-1305

https://docs.alfresco.com/content-services/latest/admin/security/#alfresco-keystore-configuration

### DIFF
--- a/roles/repository/tasks/main.yml
+++ b/roles/repository/tasks/main.yml
@@ -3,12 +3,31 @@
   include_vars: "{{ item }}"
   loop: "{{ lookup('first_found', os_fallback, errors='ignore', wantlist=True) }}"
 
-- name: Set java options for security
-  set_fact:
-    acs_secure_environment:
-      JAVA_TOOL_OPTIONS:
-        - -Dsolr.sharedSecret="{{ reposearch_shared_secret }}"
-  when: acs.version is version('7.2.0', '>=')
+- name: Handle secure properties
+  block:
+    - name: Initialize temporary accumulator
+      set_fact:
+        temp_array: []
+
+    - name: Set solr.sharedSecret property
+      set_fact:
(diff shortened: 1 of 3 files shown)


roles/repository/tasks/main.yml
```diff
@@ -3,12 +3,31 @@
...
+      set_fact:
+        temp_array: '{{ temp_array }} + [ ''-Dsolr.sharedSecret="{{ reposearch_shared_secret }}"'' ]'
+      when: acs.version is version('7.2.0', '>=')
+
+    - name: Set metadata-keystore.password and related properties
+      set_fact:
+        temp_array: '{{ temp_array }} + [
+          ''-Dmetadata-keystore.password="{{ repo_custom_keystore_password }}"'',
+          ''-Dmetadata-keystore.aliases=metadata'',
+          ''-Dmetadata-keystore.metadata.password="{{ repo_custom_keystore_password }}"'',
```
gionn, line 22: not sure if it make sense to handle all of this together since it's just how alfresco handle that
or
the user should be able to set a custom alias and provide a password for that too

gionn: @alxgomz https://docs.alfresco.com/content-services/latest/admin/security/#alfresco-keystore-configuration

looks like they should be both configurable?

alxgomz: `metadata-keystore.password` is used to open the keystore
`metadata-keystore.metadata.password` to read the key contained within the keystore (so the repo can read properties encrypted with that key.
So yeah technically these should be different. Though, while it make sense for a customer to set the keystore password I'm not sure how much we want to let him specify the password of the encryption key..
As for the alias itself I don't see a benefit in allowing people to use a different alias name

gionn: for existing installations it probably needs to be configurable by the user

PR #364, opened 2022-06-16
OPSEXP-1308 Add CSRF configuration

alxgomz:
Ref: OPSEXP-1308

Rely on the `known_urls` host variable to build CSRF configuration and document it.


roles/repository/templates/alfresco-global.properties.j2
```diff
@@ -68,6 +68,25 @@ dsync.service.uris={% if use_ssl | bool %}https{% else %}http{% endif %}://{% if
...
+{% else -%}
+  false
+{% endif -%}
+csrf.filter.origin.always={% if 'referer' in csrf.force_headers | lower %}true
+{% else -%}
+  false
+{% endif -%}
+{% else -%}
+  false
+{% endif %}
```
gionn, line 88: it's barely readable with all this nested conditionals

maybe it's better to construct a key-value structure and here in the template just foreach it like at L90

roles/repository/templates/share-config-custom.xml
```diff
@@ -76,13 +76,15 @@
...
 
          <!--
             Override and set this property with a regexp that if you have placed Share behind a proxy that
             does not rewrite the Origin header.
          -->
-         <origin></origin>
+	 <origin>{{ _csrf.compute('origin', csrf.urls)}}</origin>
       </properties>
 
       <!--
```
gionn, line 90: can you rename the file to .j2 so editors won't bother with syntax highlighting?

roles/repository/templates/alfresco-global.properties.j2
alxgomz: Doing so we would only move that templating mess in an ansible task or variable definition as we would anyway  use jinja templating.
I'm not sure that's worth the effort tbh

gionn: 🙏🏻

gionn (commented):
any docs for this feature?

gionn (approved):

PR #369, opened 2022-06-24
Fixing alfresco user for T-Router service

### LINKED ISSUE #371: Hardcoded alfresco user for T-Router service
#### Bug description
In case application user (default name is alfresco) is changed in `roles/common/vars/main.yml` for T-Router service then it is not able to start.
This fix is provided in https://github.com/Alfresco/alfresco-ansible-deployment/pull/369 

In my test case, I changed user from alfresco to alfresco_new but service still points to hardcoded alfresco user:
```shell
$ cat /etc/systemd/system/alfresco-transform-router.service | grep User
User=alfresco
```

#### Target OS
Tested on RHEL 8.5 and Ubuntu 20.04.4 LTS

#### Ansible error
```shell
RUNNING HANDLER [../roles/trouter : wait-for-aio] ************************************************
fatal: [localhost]: FAILED! => {"changed": false, "elapsed": 300, "msg": "Timeout when waiting for XXX:8090"}
```

Fikili:
In case application user is different than Alfresco, T-Router service is not able to start. 
This fix uses the application user defined in `roles/common/vars/main.yml`

Fix #371

### DIFF
--- a/roles/trouter/templates/alfresco-transform-router.service
+++ b/roles/trouter/templates/alfresco-transform-router.service
@@ -4,7 +4,7 @@ After=syslog.socket network.target
 
 [Service]
 Type=simple
-User=alfresco
+User={{ username }}
 ExecStart={{ binaries_folder }}/ats-atr.sh
 
 [Install]


gionn:
Looks a good catch, but do you mind opening also an issue with the reproduction steps and the actual error you get? That would help us to reproduce and maybe design a test case to cover that.

Fikili:
Ok, I will create the issue as well ;-)

Fikili:
@gionn The issue has been created, sorry for delay.

gionn (approved):

alxgomz:
OK given the it's a simple fix, though t-router changes should run against enterprise test in normal circumstances.

alxgomz (approved):
LGTM

PR #372, opened 2022-06-29
OPSEXP-1488 Bump versions for ACS 7.3

gionn:
OPSEXP-1488

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -2,7 +2,7 @@
 artifacts_repositories:
   enterprise:
     base_url: "https://artifacts.alfresco.com/nexus/content"
-    repository: "groups/private"
+    repository: "groups/internal"
     group_id: "org/alfresco"
   community:
     base_url: "https://artifacts.alfresco.com/nexus/content"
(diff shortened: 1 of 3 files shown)


group_vars/all.yml
```diff
@@ -2,7 +2,7 @@
 artifacts_repositories:
   enterprise:
     base_url: "https://artifacts.alfresco.com/nexus/content"
-    repository: "groups/private"
+    repository: "groups/internal"
```
gionn, line 5: we don't want this on master

gionn: because otherwise we can't test here -M versions, private doesn't contains them

gionn:
Moving fixes of tests in a separated PR so we can merge immediately https://github.com/Alfresco/alfresco-ansible-deployment/pull/386

group_vars/all.yml
alxgomz: We need an action to specifically check no reference to internal repo is merged in master

alxgomz (changes requested):
Works for me but not to merge in master.

PR #373, opened 2022-06-29
OPSEXP-1505 Fix curl issue in the env

slohe1:
curl has been added in the common roles vars.


gionn:
please remember to always reference the issue number in the PR title so we get it automatically in the commit message and we have easier life when writing release notes

PR #374, opened 2022-06-29
OPSEXP-1309: add CORS policy config

alxgomz:
Enable CORS by default and provide a documented configuration mechanism.


docs/SECURITY.md
```diff
@@ -48,4 +48,41 @@ csrf:
 
 ## Configure CORS
 
-TODO as part of OPSEXP-1309
+The playbook now enables CORS headers to be sent by the server by default. The default policy basically only allows the Same-Origin policy.
```
gionn, line 51: this has an impact only if customer has some custom application accessing alfresco or is it valid also for the web apps we manage? (adw?)

either way, maybe it's worth adding a link to this section also from the playbook-upgrade.md?

alxgomz: This is valid for any app (including ours) that's served on a different domain. At the moment the playbook do not supports deploying components on different domains so that is not a problem.

docs/SECURITY.md
```diff
@@ -20,15 +20,15 @@ all:
             - https://app.domain.local/legit
 ```
 
-Note the `known_urls` variable. It needs to be defined as a list of URLs where the client applicaiton is hosted
+Note the `known_urls` variable. It needs to be defined as a list of URLs where the client application is hosted
```
gionn, line 23:
```suggestion
Note the `known_urls` variable. It needs to be defined as a list of URLs where the client application is hosted.
```

gionn (dismissed):

PR #377, opened 2022-07-01
OPSEXP-1563 Allow overriding maven repository URL

gionn:
OPSEXP-1563

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -1,13 +1,17 @@
 ---
-common_repository:
-  base_url: "https://artifacts.alfresco.com/nexus/content"
-  repository_community: "groups/public"
-  repository_enterprise: "groups/private"
-  group_id: "org/alfresco"
+artifacts_repositories:
+  enterprise:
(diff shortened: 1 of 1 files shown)


group_vars/all.yml
```diff
@@ -1,7 +1,13 @@
 ---
+common_repository:
+  base_url: "https://artifacts.alfresco.com/nexus/content"
+  repository_community: "groups/public"
+  repository_enterprise: "groups/private"
+  group_id: "org/alfresco"
+
```
alxgomz, line 7: What about having a yaml structure as follow:

```yaml
artifacts_repositories:
  - id: enterprise
    base_url: ....
  - id: community
    base_url: .....
``` 

This would allow for customers to provide additional repos of their own by adding an entry to this list and re-use our logic to download artifacts

alxgomz (approved):
you don't like sequences do you? XD

gionn:
> you don't like sequences do you? XD

no need for an array when you have an unique key and can iterate on keys 😛

PR #379, opened 2022-07-05
OPSEXP-1296 Drop centos 8 support in ansible

himanshu-nigam:
Drop centos 8 support in ansible


group_vars/all.yml
```diff
@@ -131,4 +131,4 @@ use_ssl: false
 # Put your domain name here if any (e.g. ecm.alfresco.com). Needs to be empty if unused (don't use dummy values)
 fqdn_alfresco: ""
 
-autogen_unsecure_secrets: false
+autogen_unsecure_secrets: true
```
alxgomz, line 134: We don't want to allow unsecure setup by defauli

Vagrantfile
```diff
@@ -1,11 +1,13 @@
...
-    v.memory = 10240
-    v.cpus = 4
-  end
+config.vm.box = "ubuntu/focal64"
+config.vm.network "private_network", ip: "192.168.56.100"
+config.vm.network "forwarded_port", guest: 80, host: 80
+config.vm.provider "virtualbox" do |v|
+  v.memory = 10240
+  v.cpus = 4
+end
```
alxgomz, line 8: Is this indentation change needed?

playbooks/secrets-init.yml
```diff
@@ -47,7 +47,7 @@
           when: vault_init|default('') == 'plugin'
 
         - name: Populate secrets.yml with missing secrets
-          ansible.builtin.shell: "{{ base_folder }}/scripts/generate-secret.sh -s {{ item }} -m {{ gs_mode }} >> {{ secrets_file }}"
+          ansible.builtin.shell: "bash {{ base_folder }}/scripts/generate-secret.sh -s {{ item }} -m {{ gs_mode }} >> {{ secrets_file }}"
```
alxgomz, line 50: This has not proven to be any useful. Please revert

start
```diff
@@ -0,0 +1 @@
+vboxdrv
```
alxgomz, line 1: Can you explain what this bit is for?

alxgomz (changes requested):
There are also distribution specific files you need to get rid of.

gionn (changes requested):
PR title should contain a summary of the change (like what is currently in the description)

playbooks/secrets-init.yml
himanshu-nigam: It does not impact any changes so same work with or without bash

start
himanshu-nigam: I guess this is used to start Virtualbox service

Vagrantfile
alxgomz: Sorry I don't understand. Why does this change appears in the diff? If you did not changed it on purpose, please revert.

playbooks/secrets-init.yml
alxgomz: please be conservative and only comit changes we know had value of fix things

start
alxgomz: Again if that is not needed please revert.

alxgomz (changes requested):

PR #381, opened 2022-07-06
OPSEXP-1296 Drop support for centos8 in ansible

himanshu-nigam:
Drop support for centos8


gionn (changes requested):
there are some specific files that needs to be removed as well, like https://github.com/Alfresco/alfresco-ansible-deployment/blob/master/roles/common/vars/CentOS8.yml

himanshu-nigam:
Remove centos8.yml

alxgomz (changes requested):
postgres & transformers roles also contain var files dedicated to Centos8 that need to be removed

alxgomz (approved):
LGTM

gionn (approved):

PR #399, opened 2022-08-01
OPSEXP-1607: Enforce minimum supported ansible stack version in the playbook

kklassa:
Current implementation checks the Ansible version as well as the Jinja2 version directly in the `check-versions.yml` playbook. The main playbook `acs.yml` aborts it's execution if Ansible or Jinja2 version is older than our suggested requirements (but not if it's so old that it prevents the the playbook from even starting).  Example output for deployment using vagrant with outdated Ansible:
image


gionn (changes requested):
* for ansible version is simpler to rely on facts, see https://stackoverflow.com/a/55631911/547195
* most of the things you did in the script could be done using ansible tasks, relying on external scripts is not a good practice, try to rewrite the jinja version check as ansible task (even simpler, I will avoid doing loops if we need to check only two things)
* rewrite the description of this pr that looks outdated

kklassa:
I just discovered during tests that if I downgrade Jinja2 substantially (a major version down) the playbook doesn't even start to execute and exits with errors related to Jinja2 being too old, so although this PR in it's current state is useful for checking some inconsistencies in versioning it doesn't actually solve the most important problem and the reason behind OPSEXP-1607 in the first place. I will have to think of another solution

playbooks/check-versions.yml
```diff
@@ -0,0 +1,21 @@
...
+      fail:
+        msg: "Installed Ansible version is too old. Upgrade to at least 2.12.7"
+      when: (ansible_version.major < 2) or
+            (ansible_version.major == 2 and ansible_version.minor < 12) or
+            (ansible_version.major == 2 and ansible_version.minor == 12 and ansible_version.revision < 7)
+    - name: Check Jinja2 version
+      shell: | 
+        JINJA_VERSION=$(pip3 freeze | grep Jinja2 | sed 's/Jinja2==//g')
+        echo $JINJA_VERSION | tr '.' '\n'
+      register: version
```
gionn, line 14: use something more specific like jinja2_version because in ansible all variables are defined in a shared global context and can save a lot of headaches

playbooks/check-versions.yml
```diff
@@ -0,0 +1,21 @@
+---
+- hosts: localhost
+  tasks:
+    - name: Fail if Ansible version is too old
+      fail:
+        msg: "Installed Ansible version is too old. Upgrade to at least 2.12.7"
+      when: (ansible_version.major < 2) or
+            (ansible_version.major == 2 and ansible_version.minor < 12) or
+            (ansible_version.major == 2 and ansible_version.minor == 12 and ansible_version.revision < 7)
```
gionn, line 9: no need to check revision, 2.12.x is enough, so we don't bother user to update every time we bump patch version

playbooks/check-versions.yml
```diff
@@ -0,0 +1,21 @@
...
+    - name: Check Jinja2 version
+      shell: | 
+        JINJA_VERSION=$(pip3 freeze | grep Jinja2 | sed 's/Jinja2==//g')
+        echo $JINJA_VERSION | tr '.' '\n'
+      register: version
+    - name: Fail if Jinja2 version is too old
+      fail:
+        msg: "Installed Jinja2 version is too old. Upgrade to at least 3.1.2"
+      when: (version.stdout_lines[0] | int < 3) or
+            (version.stdout_lines[0] | int == 3 and version.stdout_lines[1] | int < 1) or
```
gionn, line 19: for jinja I think is enough to check that major is at least 3

playbooks/check-versions.yml
```diff
@@ -0,0 +1,21 @@
...
+    - name: Fail if Ansible version is too old
+      fail:
+        msg: "Installed Ansible version is too old. Upgrade to at least 2.12.7"
+      when: (ansible_version.major < 2) or
+            (ansible_version.major == 2 and ansible_version.minor < 12) or
+            (ansible_version.major == 2 and ansible_version.minor == 12 and ansible_version.revision < 7)
+    - name: Check Jinja2 version
+      shell: | 
+        JINJA_VERSION=$(pip3 freeze | grep Jinja2 | sed 's/Jinja2==//g')
+        echo $JINJA_VERSION | tr '.' '\n'
```
gionn, line 13: no need for this, just use version.split('.')[0] later

gionn (changes requested):

playbooks/check-versions.yml
```diff
@@ -0,0 +1,22 @@
...
+        JINJA_VERSION=$(pip3 freeze | grep Jinja2 | sed 's/Jinja2==//g')
+        echo $JINJA_VERSION
+      register: jinja2_version
+    - name: Fail if Jinja2 version is too old
+      fail:
+        msg: 
+          - "Installed Jinja2 major version is too old. Upgrade to at least version 3."
+          - "Please refer to https://github.com/Alfresco/alfresco-ansible-deployment/blob/master/docs/deployment-guide.md#setup-a-control-node for detailed instructions."
+      when: (jinja2_version.stdout.split('.')[0] | int < 3)
+
```
gionn, line 22:
```suggestion
```

playbooks/acs.yml
```diff
@@ -1,4 +1,7 @@
 ---
+- name: Check module requirements
```
gionn, line 2:
```suggestion
- name: Check required dependencies
```

gionn (changes requested):

gionn (approved):

PR #401, opened 2022-08-08
OPSEXP-1604 Rocky Linux 8 support

gionn:
OPSEXP-1604


roles/transformers/tasks/imagemagick-rpm.yml
```diff
@@ -0,0 +1,17 @@
...
+  vars:
+    pkg_name: "{{ item | replace('imagemagick-distribution', 'imagemagick-distribution-' + imagemagick_version) }}.{{ pkg_format }}"
+    pkg_url: "{{ nexus_repository.releases }}/../imagemagick/imagemagick-distribution/{{ imagemagick_version }}/{{ pkg_name }}"
+  get_url:
+    url: "{{ pkg_url }}"
+    checksum: "sha1:{{ lookup('url', pkg_url + '.sha1') }}"
+    dest: "{{ download_location }}/{{ pkg_name }}"
+    mode: "0644"
+  loop: "{{  imagemagick_packages }}"
+
```
gionn, line 11:
```suggestion
```

roles/transformers/vars/main.yml
```diff
@@ -9,3 +9,5 @@ libreoffice_home: "/opt/libreoffice{{ dependencies_version.libreoffice[0:3] }}/"
 
 alfresco_pdf_renderer: "{{ nexus_repository.releases }}/alfresco-pdf-renderer/{{ dependencies_version.pdf_renderer }}/alfresco-pdf-renderer-{{ dependencies_version.pdf_renderer }}-linux.tgz"
 alfresco_pdf_renderer_sha1_checksum_url: "{{ alfresco_pdf_renderer }}.sha1"
+
+imagemagick_repo: "https://github.com/Alfresco/imagemagick-build/releases/download"
```
gionn, line 13:
```suggestion
imagemagick_repo: "https://github.com/Alfresco/imagemagick-build/releases/download"

```

tests/molecule/Dockerfile-noprivs.j2
```diff
@@ -15,11 +15,12 @@ ENV {{ var }} {{ value }}
...
-    elif [ $(command -v dnf) ]; then dnf makecache && dnf --assumeyes install /usr/bin/python3 /usr/bin/python3-config /usr/bin/dnf-3 sudo bash iproute && dnf clean all; \
+    elif [ $(command -v dnf) ]; then dnf makecache && dnf --assumeyes install /usr/bin/python3 /usr/bin/python3-config /usr/bin/dnf-3 sudo bash iproute procps-ng && dnf clean all; \
     elif [ $(command -v yum) ]; then yum makecache fast && yum install -y /usr/bin/python /usr/bin/python2-config sudo yum-plugin-ovl bash iproute && sed -i 's/plugins=0/plugins=1/g' /etc/yum.conf && yum clean all; \
     elif [ $(command -v zypper) ]; then zypper refresh && zypper install -y python3 sudo bash iproute2 && zypper clean -a; \
     elif [ $(command -v apk) ]; then apk update && apk add --no-cache python3 sudo bash ca-certificates; \
     elif [ $(command -v xbps-install) ]; then xbps-install -Syu && xbps-install -y python3 sudo bash ca-certificates iproute2 && xbps-remove -O; fi
+
```
gionn, line 23:
```suggestion
```

tests/molecule/Dockerfile-noprivs.j2
```diff
@@ -28,3 +29,4 @@ RUN set -xe \
   && useradd -m -g ${DEPLOY_GROUP} ${DEPLOY_USER} \
   && usermod -aG ${SUDO_GROUP} ${DEPLOY_USER} \
   && sed -i "/^%${SUDO_GROUP}/s/ALL\$/NOPASSWD:ALL/g" /etc/sudoers
+
```
gionn, line 32:
```suggestion
```

roles/transformers/tasks/dependencies-rpm.yml
```diff
@@ -10,15 +10,16 @@
...
         creates: /opt/libreoffice{{ dependencies_version.libreoffice | regex_search('^[0-9]+.[0-9]+') }}/program/soffice.bin
         warn: false
+
+    - name: Include RockyLinux specific tasks for ImageMagick
+      include_tasks: "imagemagick-rocky.yml"
+      when: ansible_distribution == 'Rocky'
+
+    - name: Include remaining RPM distributions tasks for ImageMagick
+      include_tasks: "imagemagick-rpm.yml"
+      when: ansible_distribution != 'Rocky'
```
alxgomz, line 25: I don't like that we have splited logic here. If the reason is just the different package names or repos we should instead use per distribution variables

roles/transformers/vars/main.yml
alxgomz: Although github sounds like a better place to store than nexus I think we should still store artifact in one single place to avoid having too many dependencies.
Built packages should be uploaded to and gathered from nexus.
Also I think there are some restrictions to ImageMagick distribution which I'm not sure we conform to on this repo. At least this should be done with caution and with legal approval (https://imagemagick.org/script/license.php).
Generally speaking moving artifacts to a different place should be done in a consolidated way as part of a different ticket

roles/transformers/tasks/main.yml
```diff
@@ -29,8 +28,8 @@
     checksum: "sha1:{{ lookup('url', pkg_url + '.sha1') }}"
     dest: "{{ download_location }}/{{ pkg_name }}"
     mode: "0644"
-    timeout: 570
   loop: "{{  imagemagick_packages }}"
+  when: ansible_distribution != 'Rocky'
```
alxgomz, line 32: We should avoid spreading tasks that essentially do the same thing in different places.
Let's use distro specific vars to provide a  common interface to a single task

alxgomz (changes requested):

alxgomz (approved):
🙏 LGTM

PR #406, opened 2022-08-16
OPSEXP-1661:  Author new github action on ansible repository to avoid duplications in workflows

piraartur:
OPSEXP-1661

### DIFF
--- a/.github/actions/pipenv/action.yml
+++ b/.github/actions/pipenv/action.yml
@@ -4,9 +4,8 @@ description: 'Install python packages and create pipenv'
 runs:
   using: "composite"
   steps: 
-    steps:
-    - name: Install pipenv and Pipfile
-      shell: bash
-      run: |
-        pip install --user pipenv
-        pipenv install
+  - name: Install pipenv and Pipfile
+    shell: bash
+    run: |
+      pip install --user pipenv
+      pipenv install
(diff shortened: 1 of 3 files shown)


.github/actions/pipenv/action.yml
```diff
@@ -0,0 +1,11 @@
...
+description: 'Install python packages and create pipenv'
+
+runs:
+  using: "composite"
+  steps: 
+  - name: Install pipenv and Pipfile
+    shell: bash
+    run: |
+      pip install --user pipenv
+      pipenv install --deploy --dev
```
gionn, line 11: split in two different step so the output will be more readable

.github/actions/pipenv/action.yml
```diff
@@ -0,0 +1,11 @@
+name: install_python_packages
```
gionn, line 1:
```suggestion
name: pipenv
```

usually the name of the upper folder or a shorter description

.github/actions/pipenv/action.yml
```diff
@@ -0,0 +1,11 @@
...
+description: 'Install python packages and create pipenv'
```
gionn, line 2:
```suggestion
description: 'Install python packages with pipenv'
```

gionn (changes requested):

kklassa (approved):
looks fine in my opinion

gionn (approved):

PR #410, opened 2022-08-23
OPSEXP-1377 Support configuring jmx/rmi in sync service

gionn:
* Disable JMX by default
* Fix random startup error due to port conflicts

OPSEXP-1377

### DIFF
--- a/roles/sync/vars/main.yml
+++ b/roles/sync/vars/main.yml
@@ -1,5 +1,2 @@
 ---
 # vars file for sync
-jmx_config_remote_enabled: false
-jmx_config_remote_port: 50800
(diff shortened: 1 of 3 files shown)


roles/sync/vars/main.yml
```diff
@@ -1,2 +1,5 @@
 ---
 # vars file for sync
+jmx_config_remote_enabled: false
+jmx_config_remote_port: 50800
+jmx_config_remote_rmi_port: 50801
```
alxgomz, line 5: This should go to `defaults/main.yml`

alxgomz (changes requested):

Issue #412, opened 2022-08-24
ImageMagick download fails with SSLV3 Alert Handshake Failure


amberream:
#### Bug description

I'm trying to install Alfresco Community Edition on CentOS 7. I am using Python3.10, Ansible 2.12.4, and OpenSSL1.1.1. I get this error when it tries to download ImageMagick.

TASK [../roles/transformers : Download ImageMagick distribution] *************************
fatal: [localhost]: FAILED! => {"msg": "An unhandled exception occurred while running the lookup plugin 'url'. Error was a <class 'ansible.errors.AnsibleError'>, original message: Failed lookup url for [https://artifacts.alfresco.com/nexus/content/groups/public/org/alfresco/../imagemagick/imagemagick-distribution/7.0.10-11/imagemagick-distribution-7.0.10-11-libs-linux.rpm.sha1](https://artifacts.alfresco.com/nexus/content/groups/public/org/imagemagick/imagemagick-distribution/7.0.10-11/imagemagick-distribution-7.0.10-11-libs-linux.rpm.sha1) : <urlopen error [SSL: SSLV3_ALERT_HANDSHAKE_FAILURE] sslv3 alert handshake failure (_ssl.c:997)>. Failed lookup url for [https://artifacts.alfresco.com/nexus/content/groups/public/org/alfresco/../imagemagick/imagemagick-distribution/7.0.10-11/imagemagick-distribution-7.0.10-11-libs-linux.rpm.sha1](https://artifacts.alfresco.com/nexus/content/groups/public/org/imagemagick/imagemagick-distribution/7.0.10-11/imagemagick-distribution-7.0.10-11-libs-linux.rpm.sha1) : <urlopen error [SSL: SSLV3_ALERT_HANDSHAKE_FAILURE] sslv3 alert handshake failure (_ssl.c:997)>"}

#### Target OS

Cent OS 7

#### Ansible error

```shell
TASK [../roles/transformers : Download ImageMagick distribution] ***************
task path: /home/qoppa/Downloads/alfresco-ansible-deployment-2.0.0/roles/transformers/tasks/main.yml:24
exception during Jinja2 execution: Traceback (most recent call last):
  File "/usr/local/lib/python3.10/urllib/request.py", line 1348, in do_open
    h.request(req.get_method(), req.selector, req.data, headers,
  File "/usr/local/lib/python3.10/http/client.py", line 1282, in request
    self._send_request(method, url, body, headers, encode_chunked)
[...]

PR #414, opened 2022-08-25
try to replicate issue 412

alxgomz:
sample PR to trigger community builds for issue #412


alxgomz:
confirmed. issue tracked in OPSEXP-1698

Issue #413, opened 2022-08-25
"Ensure a list of packages installed" task failed with ansible 2.13


amberream:
#### Bug description

I'm installing alfresco community edition 7.2 on CentOS 7 with Python 3.8 and Ansible 2.13.  It fails with the following error:

TASK [../roles/repository : Ensure a list of packages installed] ***************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "It appears that a space separated string of packages was passed in as an argument. To operate on several packages, pass a comma separated string of packages or a list of packages."}

The task in roles/repository/tasks/main.yml looks like:
- name: Ensure a list of packages installed
  become: true
  package:
    name: "{{ utils_repo }} + {{ utils_storage[cs_storage.type | default('tmpfs')] }}"
    state: present

#### Target OS

CentOS 7

#### Ansible error

```shell
TASK [../roles/repository : Ensure a list of packages installed] ***************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "It appears that a space separated string of packages was passed in as an argument. To operate on several packages, pass a comma separated string of packages or a list of packages."}

```

#### Ansible context

Paste the output of the following commands:

##### ansible --version

```shell
ansible [core 2.13.3]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /root/.local/lib/python3.8/site-packages/ansible
  ansible collection location = /root/.ansible/collections:/usr/share/ansible/collections
  executable location = /root/.local/bin/ansible
  python version = 3.8.10 (default, Aug 24 2022, 18:42:19) [GCC 4.8.5 20150623 (Red Hat 4.8.5-44)]
  jinja version = 3.0.3
  libyaml = True

```

##### ansible-config dump --only-changed

```shell

```

##### ansible-inventory -i your_inventory_file --graph

```shell
@all:
  |--@activemq:
  |  |--@repository:
  |  |  |--localhost
  |--@adw:
  |  |--@repository:
  |  |  |--localhost
  |--@database:
  |  |--@repository:
  |  |  |--localhost
  |--@external:
  |  |--@external_activemq:
  |--@external_activemq:
  |--@nginx:
  |  |--@repository:
  |  |  |--localhost
  |--@repository:
  |  |--localhost
  |--@search:
  |  |--@repository:
  |  |  |--localhost
  |--@syncservice:
[...]

PR #415, opened 2022-08-25
Test with Ansible 2.13.3

gionn:
ref #413


gionn:
there are some [troubles](https://github.com/Alfresco/alfresco-ansible-deployment/runs/8017748830?check_suite_focus=true#step:11:82) with the ec2 plugin with older OSes in the integrations testing but everything else looks fine

fixes will be merged in https://github.com/Alfresco/alfresco-ansible-deployment/pull/416

PR #418, opened 2022-08-30
OPSEXP: 1372 Porting ansible actions to 2.10 fully qualified collection names

piraartur:
Creating draft pull request to test if changing true and false to suggested in documentation yes and no works
Changed some actions names to Fully Qualified Collection Name (FQCN)


piraartur:
Closing PR to move to smaller, simpler PRs instead of one big

PR #419, opened 2022-08-30
OPSEXP-1372

piraartur:


piraartur:
That's first part of changes to playbooks for fully qualified names. As in the description, I chose to move from true/false to yes/no in playbooks because  they are widely used in official ansible documentation, for example:
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/gather_facts_module.html

_A toggle that controls if the fact modules are executed in parallel or serially and in order. This can guarantee the merge order of module facts at the expense of performance.

By default it will be true if more than one fact module is used.

Choices:

no
yes_

Mieszko96 (commented):
change commit name!

rest LGMT

alxgomz (commented):
Is there a reason why these changes are limited to 2 playbooks and to the become attribute?

piraartur:
> Is there a reason why these changes are limited to 2 playbooks and to the become attribute?

From jira issue:

> Make a separated PR for each role to avoiding a single large PR.
So this PR changes two files at once, leaves 6 to made changes to and it separates changes to avoid one large PR

alxgomz:
sure @piraartur but the playbooks you started working on also contains other modules which have not been moved to fqdn names (`ansible.builtin.fail`) or included tasks file which are not part or this PR.
I'm afraid we might miss a lot of code if we don't have well defined approach.
So in addition to dealing with roles one by one, let's also make sure a playbook is fully covered before merging (meaning the playbook tasks themselves and also included files. Roles will be dealt with in separate PR).

alxgomz (approved):
LGTM

PR #427, opened 2022-09-12
OPSEXP-1708: OS check prereqs

piraartur:


group_vars/all.yml
```diff
@@ -43,6 +43,25 @@ adw:
 sync:
   version: 3.6.0
 
+supported_os:
+  RHEL:
+    versions:
+      - 8.4
```
alxgomz, line 49: Please also add
 RHEL 8.5

group_vars/all.yml
```diff
@@ -43,6 +43,25 @@ adw:
...
+      - 8.2
+      - 7.7
+      - 7.6
+  CentOS:
+    versions:
+      - 7
```
alxgomz, line 55: set it to 7.9 (which is latest CEntos anything older than that is anyway a bad idea ... w7.9. itself is no good idea :) )

alxgomz (changes requested):
nearly there

docs/deployment-guide.md
```diff
@@ -529,6 +529,9 @@ Or to deploy ACS Community use the following command:
 pipenv run ansible-playbook playbooks/acs.yml -i inventory_local.yml -e "@community-extra-vars.yml"
 ```
 
+By default, the ACS playbook will now also check compatibility of OS if it is  fully supported.
+You can add flag -e skip_os_test(true) if you want to deploy on not supported OS distribution.
```
alxgomz, line 533: paranthesis notation will cause issues in a shell. You can only use `=`

alxgomz (changes requested):

playbooks/acs.yml
```diff
@@ -1,4 +1,14 @@
...
+- name: Check if OS distribution is supported
+  hosts: all:!external
+  gather_facts: yes
+  tasks:
+  - name: compare host OS with supported matrix
+    vars:
+      os_versions: "{{ supported_os[ansible_distribution].versions | default([]) }}"
+    ansible.builtin.fail:
+      msg: "{{ ansible_distribution }} {{ ansible_distribution_major_version }} is not supported OS"
+    when: skip_os_test | default(false) and ansible_distribution_major_version | int not in os_versions
```
alxgomz, line 11: that logic is broken as `skip_os_test=true` would result in the task being executed.

```suggestion
    when:
    - not skip_os_test | default(false) | bool
    - ansible_distribution_major_version | int not in os_versions
```

group_vars/all.yml
```diff
@@ -43,6 +43,26 @@ adw:
...
+    versions:
+      - 7.9
+  Ubuntu:
+    versions:
+      - 20.04
+      - 18.04
+  SUSE:
+    versions:
+      - 15.0
+      - 12.0
```
gionn, line 64: @alxgomz not sure we could really consider suse supported since we are not testing it

also strange that it's here given that is not even mentioned in https://docs.alfresco.com/content-services/latest/support/

alxgomz: It's supported for ACS 6.2 (which we still support) but you're right e don't support this distro in the playbook. It needs to be removed.

gionn: @Artur-Piechuta you can safely remove it also in https://github.com/Alfresco/alfresco-ansible-deployment/blob/60124982dc27417d38a1054f4957240987040e2a/docs/README.md#versioning

alxgomz (approved):
LGTM

PR #431, opened 2022-09-15
OPSEXP-1639 Allow specifying a different repository for each component

kklassa:
OPSEXP-1639

### DIFF
--- a/7.0.N-extra-vars.yml
+++ b/7.0.N-extra-vars.yml
@@ -1,38 +1,51 @@
 ---
 acs:
+  repository: "{{ nexus_repository.enterprise_releases }}"
   version: 7.0.1.4
   edition: Enterprise
 
+amps:
+  aos_module:
+    repository: "{{ nexus_repository.releases }}"
+    version: 1.4.0
+  device_sync:
+    repository: "{{ nexus_repository.enterprise_releases }}"
+    version: 3.4.0
+  googledrive_repo:
+    repository: "{{ nexus_repository.enterprise_releases }}"
+    version: 3.2.1
+  googledrive_share:
+    repository: "{{ nexus_repository.releases }}"
+    version: 3.2.1
+
 api_explorer:
+  repository: "{{ nexus_repository.releases }}"
   version: 7.0.0
 
 search:
+  repository: "{{ nexus_repository.releases }}"
   version: 2.0.1.1
 
 transform:
+  repository: "{{ nexus_repository.releases }}"
   version: 2.3.10
 
 trouter:
+  repository: "{{ nexus_repository.enterprise_releases }}"
   version: 1.3.2
 
 sfs:
+  repository: "{{ nexus_repository.enterprise_releases }}"
   version: 0.13.0
 
 adw:
+  repository: "{{ nexus_repository.enterprise_releases }}"
   version: 2.1.0
 
 sync:
+  repository: "{{ nexus_repository.enterprise_releases }}"
   version: 3.4.0
 
-amps:
-  aos_module:
-    version: 1.4.0
-  device_sync:
-    version: 3.4.0
-  googledrive_repo:
-    version: 3.2.1
-  googledrive_share:
-    version: 3.2.1
 
 # this overwrites (and not merge) the same structure in group_vars/all.yml
 dependencies_version:
--- a/6.2.N-extra-vars.yml
+++ b/6.2.N-extra-vars.yml
@@ -1,25 +1,32 @@
 ---
 acs:
+  repository: "{{ nexus_repository.enterprise_releases }}"
   version: 6.2.2
   edition: Enterprise
 
-amps:  
+amps:
   aos_module:
+    repository: "{{ nexus_repository.releases }}"
     version: 1.3.1
   device_sync:
+    repository: "{{ nexus_repository.enterprise_releases }}"
     version: 3.3.3.1
   googledrive_repo:
+    repository: "{{ nexus_repository.enterprise_releases }}"
     version: 3.2.0
   googledrive_share:
     version: 3.2.0
 
 api_explorer:
+  repository: "{{ nexus_repository.releases }}"
   version: 6.3.0
 
 search:
+  repository: "{{ nexus_repository.releases }}"
   version: 1.4.3
 
 sync:
+  repository: "{{ nexus_repository.enterprise_releases }}"
   version: 3.3.3.1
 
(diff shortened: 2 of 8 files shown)


kklassa:
That's all I could think of for now, let me know if you have some suggestions about improving the current approach. I will update docs after we agree on a solution.

roles/repository/tasks/main.yml
```diff
@@ -108,7 +108,7 @@
     url_password: "{{ nexus_password }}"
   register: amp_download_result
   async: 900
-  no_log: true
+  no_log: false
```
gionn, line 111: you should probably revert all of these no_log changes

gionn (commented):
* having used dict approach instead of plain string as vars keys looks really a bad choice now (but we need to live with that until the next major release of the playbook)
* we need to add a small paragraph somewhere in the docs to have a pointer for this new feature and its basic usage

docs/components-upgrade.md
```diff
@@ -82,3 +84,61 @@ If something goes wrong with the upgrade, or if tests are not successful after u
 ### Hotfix "in-place" upgrades
 
 This process will restart tomcat on all the repository nodes and there is no guarantee one node is stopped only after the others have restarted. As a consequence a short service outage needs to be scheduled
+
+## Specifing a different component repository
```
gionn, line 88: I would say that this file is not the best one given that it reference explicitly upgrading components and this topic is more broad

I will move everything under the deployment-guide in the configure-your-deployment section

docs/components-upgrade.md
```diff
@@ -82,3 +84,61 @@ If something goes wrong with the upgrade, or if tests are not successful after u
...
+
+In case you want to use a different server/repository for a specific artifact to further customize your deployment, you can override the default URL in two ways:
+
+You can change the value of `component.repository` key for the selected component, provided that the path to your custom artifact follows the convention of the default `artifacts.alfresco.com` repository. For example to change the repository of ACS artifact you would:
```
gionn, line 92: that convention is really a standard maven repository layout, we can better reference as that

docs/components-upgrade.md
```diff
@@ -82,3 +84,61 @@ If something goes wrong with the upgrade, or if tests are not successful after u
...
+
+Edit `7.0.N-extra-vars.yml`:
```
gionn, line 94: I prefer that we reference here the group_vars/all.yml as a basic use case (that it reference the latest version).

Then at the end just state that if you are installing previous acs versions, you need to apply the same changes in the proper *-extra-vars.yml

docs/components-upgrade.md
```diff
@@ -82,3 +84,61 @@ If something goes wrong with the upgrade, or if tests are not successful after u
...
+```yaml
+amp_downloads:
+  - url: "https://your.repo.com/path/to/your/artifacts/your-alfresco-aos-module.amp"
+    sha1_checksum_url: "https://your.repo.com/path/to/your/artifacts/your-alfresco-aos-module.amp.sha1"
+    dest: "{{ content_folder }}/amps_repo/alfresco-aos-module.amp"
+```
+
+You need to check in [all.yml](../group_vars/all.yml) under which section out of `downloads`, `war_downloads` and `amp_downloads` is the specific artifact you are overriding placed.
+
+> Be careful not to override the value for `dest` key, but you must include it in the file
```
gionn, line 144: I really wonder why that dest folder is defined explicitly and cannot be evaluated dyamically (@alxgomz new issue?)

7.0.N-extra-vars.yml
```diff
@@ -1,38 +1,51 @@
...
 
-amps:
-  aos_module:
-    version: 1.4.0
-  device_sync:
-    version: 3.4.0
-  googledrive_repo:
-    version: 3.2.1
-  googledrive_share:
-    version: 3.2.1
```
gionn, line 35: why you removed those?

was the same versions of the default?

gionn (approved):

7.0.N-extra-vars.yml
kklassa: I just moved them higher in the file for some more consistency, because in all other files the `amps:` section is first right after acs

kklassa:
after modifying `check_upgrades.yml` the playbook no longer fails on unsupported upgrade in vagrant deployment:
image

roles/common/tasks/check_upgrades.yml
```diff
@@ -20,54 +20,56 @@
...
+        (
           acs.edition != installed_components.acs.edition or
           acs.version.split('.')[:3] | join('.') != installed_components.acs.version.split('.')[:3] | join('.') or
           acs.version is version(installed_components.acs.version, 'lt') or
-          amps != installed_components.amps or
-          api_explorer != installed_components.api_explorer
+          amps.aos_module.version != installed_components.amps.aos_module.version or
+          amps.device_sync.version != installed_components.amps.device_sync.version or
+          amps.googledrive_repo.version != installed_components.amps.googledrive_repo.version or
+          amps.googledrive_share.version != installed_components.amps.googledrive_share.version or
```
gionn, line 30: I am only a little unsure about checking the single amps that are not really mandatory: it will fail if user remove one amp and won't check any additional amp that we don't provide by default.

so ideally it should be something like a foreach on amps checking that version is the same or was not installed before

roles/common/tasks/check_upgrades.yml
```diff
@@ -20,54 +20,59 @@
         installed_components.acs is defined and
         installed_components.amps is defined and
```
gionn, line 21:
```suggestion
```

roles/common/tasks/check_upgrades.yml
```diff
@@ -20,54 +20,59 @@
...
-          amps != installed_components.amps or
-          api_explorer != installed_components.api_explorer
+          api_explorer.version != installed_components.api_explorer.version
         )
 
+    - name: Fail on unsupported AMPS upgrades
+      fail:
+      when: |
+          installed_components.amps is defined and
+          amps[item].version != installed_components.amps[item].version
```
gionn, line 34: you probably need something like `item in amps` otherwise you get an error if an amp is present in the system but not in the current vars?

try to reproduce by removing an amp from vars and then converge again

kklassa: I indeed reproduced something similar but after few adjustments to the checks I did it should no longer be a problem I hope

PR #432, opened 2022-09-15
OPSEXP-1635 add elasticsearch role

gionn:
OPSEXP-1635

### DIFF
--- a/roles/elasticsearch/meta/main.yml
+++ b/roles/elasticsearch/meta/main.yml
@@ -1,0 +1,53 @@
+galaxy_info:
+  author: your name
+  namespace: alfresco
+  description: your role description
+  company: your company (optional)
+
+  # If the issue tracker for your role is not on github, uncomment the
+  # next line and provide a value
+  # issue_tracker_url: http://example.com/issue/tracker
+
+  # Choose a valid license ID from https://spdx.org - some suggested licenses:
+  # - BSD-3-Clause (default)
+  # - MIT
+  # - GPL-2.0-or-later
+  # - GPL-3.0-only
+  # - Apache-2.0
+  # - CC-BY-4.0
+  license: license (GPL-2.0-or-later, MIT, etc)
+
+  min_ansible_version: 2.1
+
+  # If this a Container Enabled role, provide the minimum Ansible Container version.
+  # min_ansible_container_version:
+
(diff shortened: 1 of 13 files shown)


roles/elasticsearch/tests/inventory
```diff
@@ -0,0 +1,2 @@
+localhost
```
alxgomz, line 1: What's this `tests` folder for?

roles/elasticsearch/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
+---
```
alxgomz, line 1: these tasks require `become: true`

alxgomz (changes requested):

roles/elasticsearch/meta/main.yml
```diff
@@ -0,0 +1,25 @@
...
+  license: Apache-2.0
+
+  min_ansible_version: "2.12"
+
+  platforms:
+    - name: Ubuntu
+      versions:
+        - focal
+        - bionic
+    - name: EL
```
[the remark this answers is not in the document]
gionn: the available platforms are the ones returned by the galaxy api https://galaxy.ansible.com/api/v1/platforms/ (there is a search that returns 0 results for rocky, and for centos still return EL)

not sure it really matters what we put here if we don't publish this role on galaxy, I tried to fill everything as best as I can

roles/elasticsearch/tasks/main.yml
gionn: here I have broken the deprivileged molecule tests and noticed only today when working on the new pr, will fix

roles/elasticsearch/tests/inventory
gionn: good catch, I didn't notice before, they have been created automatically by running `molecule init role`, will try to remove them

roles/elasticsearch/meta/main.yml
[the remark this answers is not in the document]
alxgomz: good point... That doesn't really match our supported and could give the false impression it works on platforms where it actually doesn't but better stick with what galaxy knows and understand

alxgomz (approved):
LGTM

PR #448, opened 2022-09-26
OPSEXP-1736 allow better debugging experience of multimachine stage on local env

alxgomz:
Ref: OPSEXP-1736


molecule/multimachine/cleanup.yml
```diff
@@ -30,7 +30,7 @@
   hosts: localhost
   gather_facts: false
   vars:
-    broker_name: molecule-{{ lookup('env', 'BUILD_NUMBER') }}
+    broker_name: molecule-{{ lookup('env', 'BRANCH_NAME') }}-{{ lookup('env', 'BUILD_NUMBER') }}
```
gionn, line 33: I remember that not using branch_name was intentional for the usual too long limitation in the name

we could enforce via precommit or truncate branch_name to avoid that

molecule/multimachine/hosts.yml
```diff
@@ -1,10 +1 @@
...
-  children:
-    external:
-      children:
-        external_activemq:
-          hosts:
-            external_activemq_hostname:
-              activemq_username: alfresco
-              activemq_port: 61617
-              activemq_protocol: ssl
+# placeholder
```
gionn, line 1: we may want to remove and gitignore this file to avoid unintentional commit?

alxgomz: I kept a placeholder as the file is referenced in the molecule config and its absence breaks the molecule run with:
```shell
CRITICAL The source path '/Users/Alexandre.Chapellon@hyland.com/workspace/alfresco-ansible-deployment/molecule/multimachine/./hosts.yml' does not exist.
```

alxgomz: Oh just got it now 👍

gionn (approved):

kklassa (approved):

molecule/multimachine/cleanup.yml
```diff
@@ -30,7 +30,7 @@
   hosts: localhost
   gather_facts: false
   vars:
-    broker_name: molecule-{{ lookup('env', 'BUILD_NUMBER') }}
+    broker_name: molecule-{{ lookup('env', 'BRANCH_NAME')[:42] }}-{{ lookup('env', 'BUILD_NUMBER') }}
```
kklassa, line 33:
```suggestion
    broker_name: molecule-{{ lookup('ansible.builtin.env', 'BRANCH_NAME')[:42] }}-{{ lookup('ansible.builtin.env', 'BUILD_NUMBER') }}
```
I am not 100% sure, but maybe after OPSEXP-1372 this should be using FQN too?

PR #449, opened 2022-09-26
OPSEXP-1727 Ensure that python version is 3.9 within pipenv

gionn:
OPSEXP-1727


piraartur (approved):

alxgomz (commented):
what about using the version that's present in `.python-version` file so we don't have to maintain that?

gionn:
looks like the parameter is not used if the virtualenv is already created, make sure to test locally after removing manually the virtualenv

alxgomz (approved):
LGTM

PR #454, opened 2022-09-27
Opsexp 1293 imagemagick Imagemagick-7.1.0-16 version update

slohe1:
Ref: OPSEXP-1293


alxgomz (approved):
OK but we need more work to allow deploying the right IM version for the right ACS version. TBD in another task

roles/transformers/vars/Ubuntu20.yml
```diff
@@ -1,2 +1,4 @@
 ---
-imagemagick_version: 7.0.10-59
+imagemagick_version: 7.1.0-16-ci
+imagemagick_packages:
+  - imagemagick-distribution-ubuntu-20.04
```
gionn, line 4: leave a blank line at the end of each file (diff should not show the ⛔ )

gionn (changes requested):

roles/transformers/vars/Ubuntu20.yml
alxgomz: my linter usually complains about empty blank lines at the end of the yaml files.
Does yours complain about the opposite?
We need to set that in our ways of working. Any preferences gyus?

slohe1: For me it is working in both the ways. 
Currently for this added a blank line at the end.

gionn (changes requested):
if version is always the same, now it can be moved to the group_vars/all.yml

remove from Debian.yml and RedHat.yml the now unused imagemagick_packages

roles/transformers/vars/Ubuntu20.yml
gionn: will address this with https://github.com/Alfresco/alfresco-ansible-deployment/pull/456

alxgomz: 🏅

alxgomz:
@gionn this is dealt with in OPSEXP-1744

gionn (approved):

PR #457, opened 2022-09-29
OPSEXP-1705 Disable installing ghostscript by default

gionn:
OPSEXP-1705


gionn:
do we also want something to make easy to install ghostscript again?

alxgomz:
> do we also want something to make easy to install ghostscript again?

yes as mentioned in the JIRA

alxgomz (approved):

PR #459, opened 2022-09-30
Hacktoberfest: deleted not defined port variable

### LINKED ISSUE #451: Get rid of ports variable
In [port_cfg](https://github.com/Alfresco/alfresco-ansible-deployment/blob/ffd8a9438c62a7367ca0eb414af42f7366559f76/roles/common/defaults/main.yml#L43) variable there are references to a `ports` variable that is not defined anywhere and it's fine for users to just override values in ports_cfg when needed.

Remove any reference to `ports.*` and replace with plain default port value.

piraartur:
Fix #451


alxgomz (approved):
Thanks for your contribution @piraartur !

PR #464, opened 2022-10-04
Search replication 

alxgomz:
This PR aims at providing support for Solr replication with one master and possibly several slaves.
It also segregates the search role from the rest of the playbook by removing any reference to the main inventory and direct reference of variables from other roles (except the common which is a role dependency anyway)
At this stage it surely requires some improvements and maybe documentation if that's worth it but I would like to first share that approach and get feedbacks.


roles/search/defaults/main.yml
```diff
@@ -1,4 +1,15 @@
...
+search_environment:  # This will add extra vars at the end of the solr.in.sh file
   SOLR_JAVA_MEM: "-Xms1g -Xmx1g"
+
+data_folder: /opt/alfresco
+alfresco:
+  host: localhost
+  port: 80
+  port_ssl: 443
+topology: standalone
+master: "{{ ansible_play_hosts_all | first }}"
```
gionn, line 12: those vars needs a prefix to avoid conflicts (would stick to search_)

roles/search/molecule/default/converge.yml
```diff
@@ -1,10 +1,15 @@
...
-        search_environment: # This will add extra vars at the end of the solr.in.sh file
-          SOLR_JAVA_MEM: "-Xms700m -Xmx900m"
+  vars:
+    search_environment: # This will add extra vars at the end of the solr.in.sh file
+      SOLR_JAVA_MEM: "-Xms700m -Xmx900m"
+  pre_tasks:
+    - name: generate a shared secret
+      set_fact:
+        reposearch_shared_secret: >-
+          {{ lookup('ansible.builtin.password', '/dev/null chars=ascii_letters,digits') }}
```
gionn, line 11: we probably want something that would not change between reruns?

gionn (commented):
would probably add a new root molecule test for testing replication mode

roles/search/molecule/default/converge.yml
alxgomz: I don't see how that could be a problem for molecule tests

gionn: for local development I could run molecule converge multiple times, and also on the CI we could have some re-run on failure policy. This would mean having a secret change in every run that also lead to a service restart, I would try to avoid that.

I also remember we had something that automatically set all secrets in molecule, don't remember what/where it was unfortunately

roles/search/defaults/main.yml
```diff
@@ -1,5 +1,21 @@
...
+data_folder: /opt/alfresco
+alfresco:
+  host: localhost
+  port: 80
+  port_ssl: 443
+
 search_flavor: alfresco-search-services  # set to alfresco-insight-engine for IE
-search_environment:  # This will add env vars at the end of the solr.in.sh file
-  SOLR_JAVA_MEM: -Xms1g -Xmx1g
+search_version: 2.0.5.1
```
gionn, line 10: why we need a version as default?

roles/search/molecule/default/molecule.yml
```diff
@@ -16,8 +16,19 @@ platforms:
     volume_mounts:
       - "/sys/fs/cgroup:/sys/fs/cgroup:ro"
     groups:
-      - repository
-      - nginx
+      - search
+  - name: search-replica-instance
+    image: ${MOLECULE_ROLE_IMAGE:-centos:7}
```
gionn, line 21: rebase the branch, we want should have rockylinux by default everywhere now

roles/search/defaults/main.yml
alxgomz: because `search_version` is defined  as a role argument in https://github.com/Alfresco/alfresco-ansible-deployment/pull/464/files#diff-bf95b68d6c46e569cab7ad9beeaf5657df593dc27d35a53d1a25ae1d488f4be4R83
We could make it a required argument but I feel like it make more sense to have the role install a predefined version (latest)

roles/search/molecule/default/converge.yml
alxgomz: not sure why that is a problem but I changed to a static string

roles/search/templates/solrcore.properties.j2
```diff
@@ -1,8 +1,13 @@
 #
 # solrcore.properties - used in solrconfig.xml
 #
-
+{% if search_topology == 'replication' and inventory_hostname != search_master %}
+enable.alfresco.tracking=false
+alfresco.cron=* * * * * ? 1900
+alfresco.nodestate.tracker.cron=* * * * * ? 1900
```
gionn, line 7: not sure what is means

roles/search/molecule/default/molecule.yml
gionn, line 21: should be rockylinux:8?

roles/search/templates/solrcore.properties.j2
alxgomz: It's an ugly way to disable Alfresco tracking. Setting `enable.alfresco.tracking=false` SHOULD do the trick but I suspect it does really work since we introduced support for dynamic registration of shards... Hence this

gionn (approved):

PR #485, opened 2022-10-18
OPSEXP-1742 Bump detect-secrets plugin

gionn:
* update detect-secrets to latest version
* remove all pragma to allow secrets AND add them to baseline as not secrets

What I've understood is:
* every time a secret detection get triggered by pre-commit hook:
  * `detect-secrets scan --baseline .secrets.baseline` to add new detections to the baseline
  * `detect-secrets audit .secrets.baseline` to mark that detection as non-secret

OPSEXP-1742

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,15 +39,10 @@ repos:
       - id: ansible-lint
 
   - repo: https://github.com/Yelp/detect-secrets
-    rev: v1.1.0
+    rev: v1.4.0
     hooks:
       - id: detect-secrets
-        name: detect generic secrets
-        args:
-          - --baseline
-          - .secrets.baseline
-          - --exclude-secrets
-          - (dummy|admin|mp6yc0UD9e|alfresco)
+        args: ["--baseline", ".secrets.baseline"]
 
   - repo: https://github.com/awslabs/git-secrets
     rev: b9e96b3
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -109,6 +109,12 @@
     },
     {
       "path": "detect_secrets.filters.heuristic.is_templated_secret"
+    },
+    {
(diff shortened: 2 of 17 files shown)


.pre-commit-config.yaml
```diff
@@ -39,15 +39,10 @@ repos:
...
-    rev: v1.1.0
+    rev: v1.4.0
     hooks:
       - id: detect-secrets
-        name: detect generic secrets
-        args:
-          - --baseline
-          - .secrets.baseline
-          - --exclude-secrets
-          - (dummy|admin|mp6yc0UD9e|alfresco)
```
alxgomz, line 50: I would keep these in order to not be bothered IF some secret really needs to be added. That gives people an option to add secret without having to deal with secrets baseline (something to document)

.secrets.baseline
```diff
@@ -100,33 +109,225 @@
     },
     {
       "path": "detect_secrets.filters.heuristic.is_templated_secret"
-    },
-    {
-      "path": "detect_secrets.filters.regex.should_exclude_secret",
-      "pattern": [
-        "(admin|mp6yc0UD9e|alfresco)"
```
alxgomz, line 107: maybe using this is even better compared to pre-commit config arg

.pre-commit-config.yaml
gionn: I've discovered that if you update the baseline providing exclusions they become persistent into the baseline itself

.secrets.baseline
alxgomz: I'd rather we: 
* keep it for convenience
* document theese password (at least admin & alfresco) as dummy ones

alxgomz (changes requested):
let's keep the dummy password in the baseline config

alxgomz (approved):

PR #490, opened 2022-10-27
pin pipenv to 2022.10.12

gionn:
it break very bad with the yesterday release

https://github.com/Alfresco/alfresco-ansible-deployment/actions/runs/3337677442/jobs/5524326683


slohe1 (commented):
LGTM. We will need to keep eye on next release so that we can remove this release and test it with the latest one in future.

PR #491, opened 2022-10-28
OPSEXP-1803 Add Search Enteprise reindex service

gionn:
OPSEXP-1803

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -1,6 +1,5 @@
 warn_list:
   - experimental
-  - deprecations
 skip_list:
   - 'risky-file-permissions'
   - 'no-changed-when'
@@ -11,7 +10,8 @@ skip_list:
   - 'no-tabs'
   - 'empty-string-compare'
   - 'fqcn-builtins'
-
+  # Allow Jinja templating inside task and play names
+  - name[template]
 kinds:
   - yaml: "**/molecule/**/vars*.yml"
   - yaml: "**/molecule/**/hosts.yml"
--- a/docs/deployment-guide.md
+++ b/docs/deployment-guide.md
@@ -97,19 +97,20 @@ Regardless of role and connection type, a consistent folder structure is used. Y
 
 The following systemd services are deployed and can be used to stop and start Alfresco components:
 
-| Service Name                        | Purpose                                                                          |
-|:------------------------------------|:---------------------------------------------------------------------------------|
-| `activemq.service`                  | ActiveMQ Service                                                                 |
-| `postgresql-<version>.service`      | Postgresql DB Service (where `<version>` is 11 for ACS 6.2.N and 13 for ACS 7.x) |
-| `nginx.service`                     | Nginx Service                                                                    |
-| `alfresco-content.service`          | Alfresco Content Service                                                         |
-| `alfresco-search.service`           | Alfresco Search Service                                                          |
-| `alfresco-shared-fs.service`        | Alfresco Shared File Store Controller Service                                    |
-| `alfresco-sync.service`             | Alfresco Sync Service                                                            |
-| `alfresco-tengine-aio.service`      | Alfresco AIO Transform Core Engine                                               |
-| `alfresco-transform-router.service` | Alfresco Transformation Router Service                                           |
-| `elasticsearch-connector.service`   | Alfresco Search Enterprise Service                                               |
-| `elasticsearch.service`             | ElasticSearch Service                                                            |
+| Service Name                              | Purpose                                                                                 |
+|:------------------------------------------|:----------------------------------------------------------------------------------------|
+| `activemq.service`                        | ActiveMQ Service                                                                        |
+| `postgresql-<version>.service`            | Postgresql DB Service (where `<version>` is 11 for ACS 6.2.N and 13 for ACS 7.x)        |
+| `nginx.service`                           | Nginx Service                                                                           |
+| `alfresco-content.service`                | Alfresco Content Service                                                                |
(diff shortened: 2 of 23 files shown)


.ansible-lint
```diff
@@ -1,6 +1,5 @@
 warn_list:
   - experimental
-  - deprecations
```
alxgomz, line 3: 👍

roles/search_enterprise/templates/elasticsearch-cfg.j2
```diff
@@ -0,0 +1,9 @@
+Environment=ALFRESCO_SHAREDFILESTORE_BASEURL=http://{{ sfs_host }}:{{ ports_cfg.sfs.http }}/alfresco/api/-default-/private/sfs/versions/1/file
+Environment=ALFRESCO_ACCEPTED_CONTENT_MEDIA_TYPES_CACHE_BASE_URL=http://{{ trouter_host }}:{{ ports_cfg.transformers.trouter }}/transform/config
+Environment=ELASTICSEARCH_INDEXNAME=alfresco
+Environment=SPRING_ACTIVEMQ_BROKERURL=failover:({{ activemq_protocol }}://{{ activemq_host }}:{{ ports_cfg.activemq.openwire }})
```
alxgomz, line 4:
```suggestion
Environment=SPRING_ACTIVEMQ_BROKERURL=failover:({{ activemq_protocol }}://{{ activemq_host }}:{{ ports_cfg.activemq[activemq_protocol | default('openwire')] }})
```

molecule/elasticsearch/verify.yml
```diff
@@ -0,0 +1,54 @@
+- name: Verify
+  hosts: all
```
alxgomz, line 2:
```suggestion
  hosts: search_enterprise
```

molecule/elasticsearch/verify.yml
```diff
@@ -0,0 +1,54 @@
...
+    - name: "Check that {{ job_name }} is really up and running"
+      become: true
+      ansible.builtin.shell: journalctl -u {{ job_name }}
+      register: error_log
+      retries: 6
+      delay: 10
+      until: error_log.stdout | regex_search('Started LiveIndexingApp')
+
+- name: Verify Reindex before first run
+  hosts: all
```
alxgomz, line 25:
```suggestion
  hosts: search_enterprise
```

molecule/elasticsearch/verify.yml
```diff
@@ -0,0 +1,54 @@
...
+      ansible.builtin.assert:
+        that:
+          - ansible_facts.services[job_name] is defined
+          - ansible_facts.services[job_name].state == 'inactive'
+
+- name: Run the Elasticsearch Connector Reindex
+  import_playbook: ../../playbooks/search-enterprise-reindex.yml
+
+- name: Verify Reindex after first run
+  hosts: all
```
alxgomz, line 40:
```suggestion
  hosts: search_enterprise[0]
```

docs/deployment-guide.md
```diff
@@ -833,6 +834,18 @@ In that case, you can add the `cluster_keepoff` variable to one of the `reposito
...
+## Maintenance
+
+After the initial deploy, may arise the need to execute maintenance tasks that are handled via specific playbooks.
+
+### Search Enterprise Reindexing
+
+You can trigger the reindexing of existing content in Search Enterprise using a dedicated playbook:
+
+```bash
+pipenv run ansible-playbook playbooks/search-enterprise-reindex.yml -i <inventory_file>.yml
```
alxgomz, line 846: We should add a warning message that if the reindex is triggerred on an index that has already indexed content that may/will create duplicates...
Wondering if we should also add a safe guard in the service file so it's not fired up manually by an admin who may actually want to start the connector itself?

alxgomz (changes requested):

docs/deployment-guide.md
gionn: I think that this type of concerns should be documented somewhere like https://docs.alfresco.com/search-enterprise/latest/admin/ and not here

can't find anything related to duplication search in the docs

gionn: anyway I added support for specifying additional parameters for the reindex job so it's possibile to do partial reindexing as per docs

roles/repository/templates/alfresco-global.properties.j2
```diff
@@ -70,7 +70,7 @@ tika.url=${localTransform.core-aio.url}
 transform.misc.url=${localTransform.core-aio.url}
 sfs.url=http://{{ sfs_host }}:{{ ports_cfg.sfs.http }}/
 
-messaging.broker.url=failover:({{ activemq_protocol }}://{{ activemq_host }}:{{ ports_cfg.activemq.openwire }})?timeout=3000
+messaging.broker.url=failover:({{ activemq_transport }}://{{ activemq_host }}:{{ ports_cfg.activemq[activemq_protocol] }})?timeout=3000
```
alxgomz, line 73: 👍

docs/deployment-guide.md
alxgomz: I'm still concerned about someone mistakenly triggering the reindex through a `systemctl` command.
We could just check for existence of a file or var before starting that process (ofc the playbook would need to set things up when invoked on purpose).

gionn: engineers reported that running multiple times the reindex is not a problem, so we should be good to go as it is

alxgomz (approved):

PR #499, opened 2022-11-11
OPSEXP-1795 Bump community extra vars for ACS 7.3

gionn:
* update version table in docs/README.md

OPSEXP-1795

### DIFF
--- a/community-extra-vars.yml
+++ b/community-extra-vars.yml
@@ -9,9 +9,6 @@ amps:
   aos_module:
     repository: "{{ nexus_repository.releases }}/aos-module/alfresco-aos-module"
     version: 1.5.0
-  device_sync:
-    repository: "{{ nexus_repository.enterprise_releases }}/services/sync/alfresco-device-sync-repo"
-    version: 3.8.0
   googledrive_repo:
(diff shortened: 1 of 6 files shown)


community-extra-vars.yml
```diff
@@ -2,22 +2,22 @@
...
-  version: 7.2.1
+  version: 7.3.0
   edition: Community
 
 amps:
   aos_module:
     repository: "{{ nexus_repository.releases }}/aos-module/alfresco-aos-module"
-    version: 1.4.1
+    version: 1.5.0
   device_sync:
```
alxgomz, line 12: I'm not sure why we have that here. Sync Service is only available to enterprise customers. Sync do no work with community.

alxgomz (changes requested):
Also fix the molecule reference to aos 1.4 in the repo scenarios mentionned by @vprovaggi

community-extra-vars.yml
gionn: was introduced in https://github.com/Alfresco/alfresco-ansible-deployment/pull/431 by mistake I guess, I can try to fix it here

alxgomz (approved):

PR #506, opened 2022-11-16
OPSEXP-1833 Bump imagemagick package version

slohe1:
OPSEXP-1833


alxgomz (dismissed):
That's not very sustainable.
make the suffix part of the version variable, or - if not possible - create a new suffix variable to append in var files.
That way we only have one place where we need to fix things in case we update that release build number

slohe1:
> That's not very sustainable. make the suffix part of the version variable, or - if not possible - create a new suffix variable to append in var files. That way we only have one place where we need to fix things in case we update that release build number

Bumped a new variable "imagemagick_release" under dependencies_version so that its one place for both imagemagick version & release number

roles/transformers/vars/CentOS7.yml
```diff
@@ -1,5 +1,5 @@
 ---
-imagemagick_version: "{{ dependencies_version.imagemagick }}-ci"
+imagemagick_version: "{{ dependencies_version.imagemagick }}-ci-{{ dependencies_version.imagemagick_release }}"
```
gionn, line 2: I still think that just using`dependencies_version.imagemagick` would be simpler and gives more flexibility on the package version

gionn (changes requested):
looking at the test failures it seems we have a problem with RHEL 8, `test_get_rendition` is always failing on integration tests

then it seems there is a problem running `libreoffice` in the transformers role tests [link](https://github.com/Alfresco/alfresco-ansible-deployment/actions/runs/3487299672/jobs/5834772497#step:7:1667)

gionn (approved):

PR #520, opened 2022-12-12
OPSEXP-1891 Switch tests to rockylinux 8

gionn:
Get rid of unreliable centos:7 docker tests

OPSEXP-1891

### DIFF
--- a/.github/workflows/community.yml
+++ b/.github/workflows/community.yml
@@ -19,7 +19,6 @@ jobs:
       fail-fast: true
       matrix:
         molecule_distro:
-          - image: centos:7
           - image: ubuntu:20.04
           - image: rockylinux:8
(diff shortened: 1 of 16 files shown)


alxgomz (changes requested):
I agree we can remove centos7 in all docker integration but as it is also used in acs6.2 EC2 tests, and we're going to remove it, I'd rather that we replace the acs7.1 test which are done with RHEL8 with centos7 (it's still supported up to 7.2 at least)

gionn:
> it is also used in acs6.2 EC2 tests, and we're going to remove it

plan was not to remove it but switch to ACS 7.0 (or 7.1?), so here should be good to go

alxgomz (approved):

PR #545, opened 2023-02-08
OPSEXP-1980: Add support for importing certificates in the JRE

alxgomz:
Ref: OPSEXP-1980


roles/java/meta/argument_specs.yml
```diff
@@ -0,0 +1,88 @@
...
+          path:
+            description: |
+              Local path to the java keystore
+            required: true
+            type: str
+          pass:
+            description: |
+              Password to unlock the java keystore
+            required: true
+            type: str
```
gionn, line 18: not sure we could have this as a role argument, how it fits with the pre-existing ansible-vault feat?

roles/java/meta/argument_specs.yml
```diff
@@ -0,0 +1,88 @@
...
+              path:
+                type: str
+                description: |
+                  Local path to the host specific PKCS12 certificate container
+                required: true
+              pass:
+                type: str
+                description: |
+                  Certificate container password
+                required: true
```
gionn, line 40: 🤔

roles/java/meta/argument_specs.yml
```diff
@@ -0,0 +1,88 @@
...
+              name:
+                type: str
+                description: |
+                  Name of the security key to  use as an alias in the keystore
+                required: true
+              pass:
+                type: str
+                description: |
+                  Passphrase of the security key
+                required: true
```
gionn, line 62: 🤔

roles/java/meta/argument_specs.yml
```diff
@@ -0,0 +1,88 @@
...
+        required: false
+        description: |
+          Local path to the java truststore
+        default: "{{ java_home }}/lib/security/cacerts"
+      java_truststore_pass:
+        type: str
+        required: false
+        description: |
+          Password to unlock the java truststore
+        default: changeit
```
gionn, line 88: 🤔

roles/java/meta/argument_specs.yml
```diff
@@ -0,0 +1,88 @@
...
+  main:
+    short_description: Main entrypoint for the java role
+    options:
+      java_keystore:
+        type: dict
+        options:
+          path:
+            description: |
+              Local path to the java keystore
+            required: true
```
gionn, line 12: probably not, if user doesn't requires mtls

gionn (commented):
I need more docs to better review, I am unsure how this will fit with the main playbook

roles/java/meta/argument_specs.yml
alxgomz: What I had in mind was:
* a we add a new secret in the `Populate secrets.yml with missing secrets` task (e.g.  `keystore_password`)
* pass that var as a role argument when calling the java role (`repository` molecule converge example below).

```yaml
- name: Converge
  hosts: all
  roles:
    - role: postgres
    - role: java
      java_keystore:
        type: JKS
        path: "{{ content_folder }}/keystore/alfresco.jks"
        pass: "{{ keystore_password }}"
        cert_containers:
          - path: "{{ inventory_hostname }}.p12"
            pass: "{{ keystore_password }}"
            trustcachain: true
    - role: activemq
    - role: repository
```

roles/java/meta/argument_specs.yml
alxgomz: This use case was declined in https://alfresco.atlassian.net/wiki/spaces/TECH/pages/1124696953/Certificate+deployment+in+Ansible+playbook

roles/java/meta/argument_specs.yml
alxgomz: p12 are always protected with passwords

roles/java/meta/argument_specs.yml
alxgomz: for compatibility purpose with the default JDK provided truststore

alxgomz:
@gionn I tried addressing your comments and added a simple readme to the role. Let me know if that is sufficient.

roles/java/meta/argument_specs.yml
```diff
@@ -0,0 +1,88 @@
...
+                required: true
+              pass:
+                type: str
+                description: |
+                  Certificate container password
+                required: true
+              trustcachain:
+                type: bool
+                description: |
+                  Whether to look for and trust CA chain
```
gionn, line 44: what about something more explicit like `add_to_trusted_ca`?

gionn (approved):

PR #548, opened 2023-02-20
OPSEXP-1982: implement hosts' certificate generation

alxgomz:
Ref: OPSEXP-1982


docs/SECURITY.md
```diff
@@ -108,12 +127,92 @@ all:
...
+## Hosts certificates & keys
+
+By default the playbook will generate a dedicated host certificate and private
+key pair. It is possible to either use your own private PKI or let the playbook
+generate a small PKI infrastructure.
+If you want to leverage your own PKI infrastructure, you need to pass the
+playbook 2 assets:
+
+- A CA signing certificate in PEM format
+- The encrypted CA signing key in PEM format
```
gionn, line 151: I am not sure this is really a common use case, most of the times the PKI is managed externally and who manage alfresco installs may not have access to CA pkey but just have a bunch of certificates generated externally (manually or not) after a pkey has been generated and a CSR submitted.

I understand that this approach is almost free with the implementation of a local PKI, but still may pretty uncommon.

docs/SECURITY.md
```diff
@@ -108,12 +127,92 @@ all:
...
+- ca_signing_key_passphrase: the passphrase to decrypt the CA signing key
+- certs_p12_passphrase: the passphrase the playbook will use to protect the
+  resulting PKCS12 containers.
+
+In order to set secrets in the playbook's vault please refer to
+[SECRETS.md](SECRETS.md).
+
+If you want to fully automate this generation just make sure to run the
+`secrets-init.yml` playbook as described in the
+[Deployment guide](./deployment-guide.md#Encrypted variables).
```
gionn, line 178:
```suggestion
[Deployment guide](./deployment-guide.md#Encrypted variables).
```

check the anchor because it's a lowercase slug without spaces

playbooks/pki.yml
```diff
@@ -0,0 +1,203 @@
...
+          register: ca_key_valid
+
+        - name: Set CA variables
+          when:
+            - ca_cert_valid.valid_at.tomorrow
+            - ca_key_valid.can_parse_key
+          set_fact:
+            ca_cert_path: "{{ ca_cert_stat.stat.path }}"
+            ca_key_path: "{{ ca_key_stat.stat.path }}"
+      rescue:
```
gionn, line 81: mmm, falling back to self-signed when externally provided pki fails doesn't seems too user friendly: if I want to use my CA but I fail to setup certs in the proper paths I endup with a success play but with a result I was not expecting.

Can we have a dedicate variabile that the user can toggle to declare its desired pki? (self-signed or not)
in that way you would trigger a failure if necessary files are not where they are expected to be

playbooks/pki.yml
```diff
@@ -0,0 +1,203 @@
...
+    - name: Generate PKCS12 keystores
+      community.crypto.openssl_pkcs12:
+        action: export
+        path: "{{ hostvars.localhost.pki_dir }}/certificates/{{ inventory_hostname }}.p12"
+        passphrase: "{{ p12_passphrase }}"
+        friendly_name: "{{ inventory_hostname }}"
+        privatekey_content: "{{ srvkey.privatekey }}"
+        certificate_path: "{{ hostvars.localhost.pki_dir }}/certificates/{{ inventory_hostname }}.crt"
+        ca_certificates: "{{ hostvars.localhost.ca_cert_path }}"
+      delegate_to: localhost
```
gionn, line 203: can you add a new docker test in the root that runs this playbook and test that the expected files are where they are supposed to be?

docs/SECURITY.md
alxgomz: I think you're talking about a different use-case. When the pki is managed entirely externally (can still be internal to customer's org ofc) it is to possible for someone to generate its own certificates. And as a consequence the only thing one can get is the list of certificates. That's a tedious approach as for people managing the PKI but no generation is required at teh playbook level in this case (all certs are considered prereqs).

playbooks/pki.yml
alxgomz: I prefer not to add more required variables if we can avoid. However I agree we can simply fail in case we find a user's CA and we can't use it.
In case there is no user CA, then we go with our selfsigned one. wdyt?

docs/SECURITY.md
gionn: ok in that case we could assume that all the certs will be placed in the same locations that the new playbook is handling 👍🏻

playbooks/pki.yml
gionn: ok so at least the CA should be put in the proper path and that's the trigger to fail if every other step would fail

docs/SECURITY.md
alxgomz: will document that.

gionn (approved):
🚀

PR #558, opened 2023-03-13
OPSEXP-2036: fix CORS & CSRF config

alxgomz:
Ref: OPSEXP-2036

`cors.allowed.origins` was not using the right separator
`known_urls` containing duplicate entries resulted in duplicate items in the regex
Share csrf configuration disable switch was not really working


roles/repository/templates/share-config-custom.xml.j2
```diff
@@ -263,7 +264,9 @@
             </action>
          </rule>
       </filter>
-
+{% else %}
+      <filter/>
```
gionn, line 268: where this filter has been opened?

roles/repository/templates/share-config-custom.xml.j2
```diff
@@ -58,7 +58,8 @@
          </action>
       </rule>
    -->
-   <config evaluator="string-compare" condition="CSRFPolicy" replace="{{ csrf_config.enabled }}">
+   <config evaluator="string-compare" condition="CSRFPolicy" replace="true">
```
gionn, line 61: is this documented somewhere? can't find it in the alfresco docs

alxgomz: It's documented in the code:
https://github.com/Alfresco/share/blob/master/packaging/distribution/src/main/resources/web-extension-samples/share-config-custom.xml#L39

roles/repository/templates/share-config-custom.xml.j2
gionn: XML 🤦🏻

gionn (commented):
the best would be to have a test to check the content of this template after ansible play

roles/repository/molecule/default/tests/test_repository.py
```diff
@@ -186,6 +203,24 @@ def test_newly_added_properties_are_set(host):
...
+                              share_cfg_dir,
+                              'share-config-custom.xml'])
+    xmlpath = './config[@condition="CSRFPolicy"]/properties'
+    with host.sudo():
+        share_cfg = host.file(share_cfg_file).content_string
+    share_xml = xmlreader.fromstring(share_cfg)
+    referer = share_xml.findall('{}/referer'.format(xmlpath))
+    origin = share_xml.findall('{}/origin'.format(xmlpath))
+    assert_that(referer[0].text == 'http://localhost:8080/share/.*')
+    assert_that(origin[0].text == 'http://localhost:8080')
```
gionn, line 226: can you add an additional url in the moelcule test configuration so we can assert that also the separators are handled correctly (since they were the actual bug for this issue)

gionn (approved):

PR #559, opened 2023-03-17
Fix OS test check and update supported OSes configuration

morgan-patou:
**Description:**
Task with name "- name: Compare host OS with supported matrix" is currently being skipped most of the time. This is because "skip_os_test" is undefined and because "ansible_distribution_version | int not in os_versions" is failing as well. The transformation to "INT" means that only the major version is kept (E.g.: RedHat 8.6 >> 8). Since "8" is never present in "os_versions" (which contains 8.2, 8.5, 8.6, etc.), then this check is always skipped. According to the documentation, it is supposed to be the opposite, meaning that it's always executed, except if "-i skip_os_test=true" is added as extra var.

Also, the correct ansible distribution for RedHat is "RedHat" and not "RHEL". This causes the "supported_os" to not find anything in case the target OS is RedHat, c.f. https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_conditionals.html#ansible-facts-distribution.

**Solution:**
This PR includes the switch from "RHEL" to "RedHat" in the group_vars/all.yml file as well as the fix of the conditions so that the check is always executed and gives the expected result, meaning:
1. failure if distribution/version doesn't match what is defined in the group_vars/all.yml file
2. skipping if distribution/version matches OR if skip_os_test=true

**Please review**

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -71,7 +71,7 @@ sync:
     {{ nexus_repository.enterprise_releases }}/services/sync/sync-dist-6.x
   version: 3.8.0
 supported_os:
-  RHEL:
+  RedHat:
     versions:
       - 8.6
       - 8.5
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -79,9 +79,14 @@ supported_os:
       - 8.2
       - 7.7
       - 7.6
+  Rocky:
+    versions:
+      - 8.7
+      - 8.6
   CentOS:
     versions:
       - 7.9
+      - 7.7
   Ubuntu:
     versions:
       - 22.04
--- a/playbooks/acs.yml
+++ b/playbooks/acs.yml
@@ -10,7 +10,7 @@
       msg:
         - "{{ ansible_distribution }} {{ ansible_distribution_version }} is not a supported OS"
     when:
-       - not (skip_os_test | default(false) | bool)
+      - not (skip_os_test | default(false) | bool)
       - ansible_distribution_version | float not in os_versions
 
 - name: Run preliminary checks
--- a/playbooks/acs.yml
+++ b/playbooks/acs.yml
@@ -10,8 +10,8 @@
       msg:
         - "{{ ansible_distribution }} {{ ansible_distribution_version }} is not a supported OS"
     when:
-      - not skip_os_test | default(true) | bool
-      - ansible_distribution_version | int not in os_versions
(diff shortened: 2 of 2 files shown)


alxgomz (changes requested):
Well spotted! Thanks for the contribution. Just one small suggestion, but that's all OK to move it forward.

morgan-patou:
Works for me with this alternative definition. I quickly tested the different possibilities (supported, unsupported, without and with false/true for the skip_os_test), no problems spotted.

gionn (commented):
There are a few test failures that expose that we missed updating the supported os matrix due to the bug you just fixed.

Rocky 8.6 and 8.7 should be added for sure, not sure about CentOS 7.7 that is an old version and I suppose we need to bump the AMI we are testing against.

up to you if you want to try to tackle also this last issues in order to obtain a green build, or just put this PR on hold and I can try to take a look soon and raise another PR to fix those tests.

morgan-patou:
> There are a few test failures that expose that we missed updating the supported os matrix due to the bug you just fixed.
> 
> Rocky 8.6 and 8.7 should be added for sure, not sure about CentOS 7.7 that is an old version and I suppose we need to bump the AMI we are testing against.
> 
> up to you if you want to try to tackle also this last issues in order to obtain a green build, or just put this PR on hold and I can try to take a look soon and raise another PR to fix those tests (internally tracked as [OPSEXP-2047](https://alfresco.atlassian.net/browse/OPSEXP-2047)).

If needed I can modify the group_vars/all.yml to include Rocky and CentOS versions mentioned but I do not know if you have any other steps to perform to include a new OS? Do you have some guidelines with the list of all steps? I would also need some details on which ACS versions are supported on these 2 OS, so I can update the doc as well. Currently CentOS 7.7 is marked as supported for all ACS versions from 6.2 to 7.3, including Community as well.

gionn:
> If needed I can modify the group_vars/all.yml to include Rocky and CentOS versions mentioned but I do not know if you have any other steps to perform to include a new OS?

You should just really need to add Rocky 8.6, Rocky 8.7, CentOS 7.7

docs should be fine https://github.com/Alfresco/alfresco-ansible-deployment/blob/master/docs/README.md#versioning

thanks again!

gionn (approved):

alxgomz (approved):
OK to merge as-is for now but we'll need another task to update our centos AMI and drop the support for centos 7.7 as most ACS version actually specify 7.9

PR #565, opened 2023-03-29
OPSEXP-1990: refactor zip builder script

alxgomz:
Ref: OPSEXP-1990

Tested build script in: https://github.com/Alfresco/alfresco-ansible-deployment/actions/runs/4556555199/jobs/8037061270

### DIFF
--- a/scripts/generate-zip.sh
+++ b/scripts/generate-zip.sh
@@ -18,13 +18,14 @@ echo "# Run the helper scripts/generate-secrets.sh to kick start your Ansible va
 
 echo "Generating ZIP file..."
 find "$ARTIFACT_NAME" \( -type f \
-	! -path './.*' \
-	! -path './tests/*' \
-	! -path '*/molecule/*' \
-	! -name '*.puml' \
-	! -path './scripts/*' \
+	! -path "./${ARTIFACT_NAME}/.*" \
+	! -path "./${ARTIFACT_NAME}/tests/*" \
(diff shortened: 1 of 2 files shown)


gionn (changes requested):
[test run](https://github.com/Alfresco/alfresco-ansible-deployment/actions/runs/4556020022/jobs/8035833653#step:5:567) doesn't look good

can you also add a workflow_dispatch, have a fallback version and skip the maven upload? so we have a consistent way to test it without any major downside

alxgomz:
> 

I linked the wrong build that is actually fixed (PR desc updated). Also added the workflow dispatch

slohe1 (dismissed):

gionn (changes requested):
.envrc can be removed

scripts/generate-zip.sh
```diff
@@ -20,12 +20,11 @@ echo "Generating ZIP file..."
 find "$ARTIFACT_NAME" \( -type f \
 	! -path "${ARTIFACT_NAME}/.*" \
 	! -path "${ARTIFACT_NAME}/tests/*" \
-	! -path "${ARTIFACT_NAME}/molecule/*" \
+	! -path "${ARTIFACT_NAME}/*/molecule/*" \
```
gionn, line 23: that will not match anymore the molecule in the root

gionn (changes requested):

PR #566, opened 2023-03-31
Move group_vars inside playbooks directory

morgan-patou:
**Description:**
Currently the `group_vars` folder is at the top level of the repository. This is fine when executing the `ansible-playbook` locally inside this folder because the `inventory` is stored inside this folder as well and therefore loads the `group_vars` that is at this exact location.
However, in case the `ansible-playbook` command has to be executed from outside with an `inventory` not in this local folder, then that means that the `group_vars` will not be loaded automatically and it needs to be loaded through the `extra-vars`.

**Possible solution:**
Move the `group_vars` from the root of the repository to inside of the `playbooks` folder. When the `group_vars` is inside the `playbooks` folder, then it should be loaded automatically for any `ansible-playbook` execution. Hence, it should still continue to behave in the same way for the standard execution inside the local folder but it should also simplify the execution from the outside, since there wouldn't be the need to load the files through the `extra-vars` anymore if the `group_vars` is inside the `playbooks` folder.

Please let me know if you would be willing to make this change. If yes, I can look at all documentations, to make sure it reflects the new path. For now, I only put the move of the folder in this PR.


alxgomz:
I'm not sure we really want to do thatas this location is only valied for the `ansible-playbook command`. General approach is to have `group_vars` & `host_vars` live together with the inventory (whether it is a file, a folder or dynamic plugin).
As we ship the example inventories at the root of the project their vars folders live there too.
If someone needs to move his inventory to somewhere else I would adice he moves the `group_vars` & `host_vars` as well.

morgan-patou:
> I'm not sure we really want to do that as this location is only valid for the `ansible-playbook command`. General approach is to have `group_vars` & `host_vars` live together with the inventory (whether it is a file, a folder or dynamic plugin). As we ship the example inventories at the root of the project their vars folders live there too. If someone needs to move his inventory to somewhere else I would advise he moves the `group_vars` & `host_vars` as well.

I'm not sure I get the comment about this being only valid for the `ansible-playbook` command, what would be other uses of the group_vars? In terms of variable inclusion, c.f. [variable precedence](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_variables.html#understanding-variable-precedence), it should be exactly the same. So it has no impact on your behavior and it would help external integrations.

alxgomz:
First on that comes to mind:
```shell
ansible-inventory -i inventory_ssh.yml --list --vars
```

morgan-patou:
> First one that comes to mind:
> 
> ```shell
> ansible-inventory -i inventory_ssh.yml --list --vars
> ```

I see. In this case since you are almost only using the all.yml (and considering it would only be for the latest enterprise version and not considering older/community params), listing vars through the inventory doesn't make much sense to me, but I guess it could be a topic at some point. If you don't want to move the group_vars, feel free to close this PR, that's why I only moved the folder for now, since I assumed it might not be approved anyway.

alxgomz:
Hi @morgan-patou ,

Thank you for the feedback. I'll go ahead and close the PR. In the future I think the best approach would be to remove as many of these vars as possible to something else than the `all` group's vars + extra-vars to something that's more consistent. We actually have a related ticket open for quite some time but I'm not sure it's accessible to you: OPSEXP-1842

PR #571, opened 2023-04-05
OPSEXP-2066 refactor amps updates in updatecli

gionn:
Ref: OPSEXP-2066

PR #576, opened 2023-04-12
OPSEXP-2066 Get rid of outdated AGS amps in tests

gionn:
Ref: OPSEXP-2066

After https://github.com/Alfresco/alfresco-ansible-deployment/pull/571 we realized that we were using very old AMPs since ages and in the meantime AGS packaging has [changed](https://docs.alfresco.com/governance-services/latest/install/zip/) and we need proper support for that.


slohe1 (commented):
Can't we remove ```MOLECULE_IT_EXTRA_VARS: tests/test-extra-vars.yml``` from other vars files, as this file is any how empty.

gionn:
> Can't we remove `MOLECULE_IT_EXTRA_VARS: tests/test-extra-vars.yml` from other vars files, as this file is any how empty.

I was thinking that we could make the variable optional (there are some tests that are using some other custom extra vars) but I kept it in case we need to toggle something on all the tests - but I did that just because it was easier

slohe1 (approved):

alxgomz (approved):

PR #577, opened 2023-04-12
OPSEXP-1956: add configuration for mTLS auth between repo and tengine (community setup)

alxgomz:
Ref: OPSEXP-1956

Leverage roles' entrypoints in java role to specifically deploy keystores rather than making it a conditional task inclusion (offers better argument validation)
Also fixes a missing dependency in transformers (exiftools)


group_vars/all.yml
```diff
@@ -19,12 +19,12 @@ nexus_repository:
 acs:
   artifact_name: alfresco-content-services-distribution
   edition: Enterprise
-  repository: "{{ nexus_repository.enterprise_releases }}"
-  version: 7.4.0-M3
+  repository: "{{ nexus_repository.development_releases }}"
```
gionn, line 22: still need to find a solution for this - and avoid merging this on master

for now maybe add a # FIXME

molecule/pki/host_vars/tls-ats-instance.yaml
```diff
@@ -0,0 +1,10 @@
+# placeholder
+ansible_user: ansible
+known_urls:
+  - http://localhost:8080/share
+  - https://someotherapp/
+acs:
+  artifact_name: alfresco-content-services-community-distribution
+  edition: Community
+  repository: "{{ nexus_repository.development_releases }}"
+  version: 7.4.0-A27
```
gionn, line 10: we should avoid having versions in molecule tests or we will continue to forget about them - why is necessary here?

roles/transformers/tasks/main.yml
```diff
@@ -8,6 +8,31 @@
...
+  yum_repository:
+    name: epel
+    description: EPEL YUM repo
+    metalink: >-
+      {{ epel_meta_base }}?repo=epel-{{ epel_ver }}{{ epel_meta_args }}
+    gpgcheck: true
+    gpgkey: >-
+      {{ epel_gpg_base }}/RPM-GPG-KEY-EPEL-{{ epel_ver }}
+  when: pkg_format == 'rpm'
+
```
gionn, line 35: move to common role?

gionn (changes requested):
I see cherry-picked commits from other PR, that means this PR should get merged AFTER them

molecule/pki/host_vars/tls-ats-instance.yaml
alxgomz: leftovers. I had this file previously during development as I thought I would make this scenario a community one. I gave up with this idea. Will remove.

gionn (approved):

PR #579, opened 2023-04-20
Test against ACS A29

gionn:


roles/repository/molecule/default/tests/test_repository.py
```diff
@@ -107,7 +107,7 @@ def test_share_log_exists(host):
 
 def test_share_context_200(host):
     "Check that /share context is available and returns a HTTP 200 status code"
-    cmd = host.run("curl -iL --connect-timeout 5 http://{}:8080/share/page/".format(test_host))
+    cmd = host.run("curl -iL --user admin:admin --connect-timeout 5 http://{}:8080/share".format(test_host))
```
kavitshah-gl, line 110: Do we need to add this? With Version A29 it should work as it was working before.

gionn: it's the revert of the test change that was introduced to have a green build with the A27 tag

alxgomz: I would revert the question.
Is it meaningful  to test that login page using a basic auth?
I changed it as i though it was not. But I have to admit that this allowed us to find  the issue you guys fixed...
If we know what we aretesting here in particular when using basic auth we can re-introduce that test (in addition to the one without auth)

alxgomz: That question is for you @kavitshah-gl  ofc :)

kavitshah-gl: Basically, the issue we identified was respect to Cookie, it was not creating a new session if the one is not present. When we looked at the code we don't see the need of passing the credentials as part of sending the GET request for "/Share". We tested the same over postman as well and it works without sending credentials.

alxgomz: Does this means that similar issue should now be surfaced even without using authentication in the future?

kavitshah-gl: Now the issue will not occur. The actual issue was nothing to do with credentials, but the Session cookie. Now that is fixed and when I tested using Postman with and without credentials, it did not give the error.

alxgomz: Sure, that I understand but my question is: If another issue with session handling happens will it be caught by a call without the the credentials or does it need to be given credentials to ensure better coverage?

kavitshah-gl: I guess it is logical to send the credentials as it was earlier. By doing so, if any issue occurs in future we make sure that it was working before (making any Keycloak Adapter replacement changes).

alxgomz: OK I will add that test back to test sessions in another PR. This one can be closed.

PR #582, opened 2023-04-25
OPSEXP-1953: Enabled sfs mTLS config

alxgomz:
Ref: OPSEXP-1953


molecule/pki/host_vars/tls-ats-instance.yaml
```diff
@@ -6,7 +6,7 @@ known_urls:
 # We are forced to use community now otherwise other transformers may conflict with mtls config
 # Remove when mTLS is implemented for SFS and trouter as well.
 acs:
-  artifact_name: alfresco-content-services-community-distribution
-  edition: Community
+  artifact_name: alfresco-content-services-distribution
+  edition: Enterprise
```
gionn, line 10: if we are back to enterprise you should not need anymore this override

molecule/pki/verify.yml
```diff
@@ -46,12 +46,18 @@
 
     - name: Check in logs a client did connect
       ansible.builtin.slurp:
-        src: /var/log/alfresco/ats-ate-aio.log
-      register: aio_log
+        src: /var/log/alfresco/{{ item.file }}.log
+      register: ats_log
```
gionn, line 50: static register + loop doesn't seems a good idea

roles/sfs/templates/ats-shared-fs.sh.j2
```diff
@@ -8,4 +8,7 @@ fi
 {% for key, value in sfs_environment.items() %}
 export {{ key }}="{{ value | join(' ') }}"
 {% endfor %}
-${JAVA_HOME}/bin/java ${JAVA_OPTS} -jar ${ATS_HOME}/alfresco-shared-file-store-*.jar > {{ logs_folder }}/ats-shared-fs.log
+${JAVA_HOME}/bin/java ${JAVA_OPTS} \
+	-jar ${ATS_HOME}/alfresco-shared-file-store-*.jar \
+	--spring.config.additional-location=optional:sfs-mtls.properties \
+	> {{ logs_folder }}/ats-shared-fs.log
```
gionn, line 14: I think there is some systemd feature to redirect logs to a specific file, so we can also add an exec prefix to the command and handle termination better, right now is probably getting killed abruptly after a stop timeout

molecule/pki/host_vars/tls-ats-instance.yaml
alxgomz: And that was even written in the comment above 🤦

molecule/pki/verify.yml
alxgomz: why not? The tests I made locally seems to prove good results as long as you realise it register the current iteration rather than the list of iterations (which you get once out of the loop)

roles/sfs/templates/ats-shared-fs.sh.j2
alxgomz: created OPSEXP-2068

molecule/pki/verify.yml
gionn: oh ok it's used only during the loop and not after - so it make sense

Issue #585, opened 2023-04-28
Verify accuracy of documentation statement


morgan-patou:
#### Description

Hello, could you please help to check the accuracy of [deployment-guide.md line 267](https://github.com/Alfresco/alfresco-ansible-deployment/blob/7764af09c54162150636679f5c290fc2abefa8d2/docs/deployment-guide.md?plain=1#L267) as well as [deployment-guide.md line 271](https://github.com/Alfresco/alfresco-ansible-deployment/blob/7764af09c54162150636679f5c290fc2abefa8d2/docs/deployment-guide.md?plain=1#L271)?

line 267: it's written port 80 for the ADW, shouldn't this be 8880 instead? (or maybe it should be removed completely since ADW needs to be on the Nginx host at the moment (unless this was fixed/changed?) and therefore communications would only happen locally from Nginx to 8880 port of ADW).

line 271: the playbook `prerequisite-checks.yml` is currently trying to connect from the repository host to the SFS / TRouter ports but according to the doc, these 2 ports are only used locally. Do you know what would be the correct statement? I assume that the 3 ports should be accessible from the Repository since we configure these details on the `alfresco-global.properties` but I'm not certain.

I'm currently writing some code to handle the firewall configuration and would appreciate some help with these details, to finish the checks and possibly fix the missing elements in the doc/existing playbooks as well.

#### Target OS

All

#### Ansible error

No specific errors, just some documentation and potentially some playbook adaptation needed depending on the correct statement.

Thanks for the check and the feedback!

gionn:
> line 267: it's written port 80 for the ADW, shouldn't this be 8880 instead? (or maybe it should be removed completely since ADW needs to be on the Nginx host at the moment (unless this was fixed/changed?) and therefore communications would only happen locally from Nginx to 8880 port of ADW).

you are right, adw role expose 8880 and expect to be reachable there by nginx node.

I am not aware on any issues when nginx and adw are not on the same node, but I know that a common deployment choice is to have them on the same node.

> line 271: the playbook `prerequisite-checks.yml` is currently trying to connect from the repository host to the SFS / TRouter ports but according to the doc, these 2 ports are only used locally. Do you know what would be the correct statement? I assume that the 3 ports should be accessible from the Repository since we configure these details on the `alfresco-global.properties` but I'm not certain.

I am not sure at this point how much we can trust that playbook given that is not part of any integration tests and may be a good idea to hook it at the end. I will give it a try.

[...]

PR #589, opened 2023-05-10
Add firewall setup and configuration

morgan-patou:
Hello team,

I spent some time to try to add the firewall configuration. I added support for both `RedHat` and `Debian` family using their default firewall (`firewalld` for RedHat / `ufw` for Debian). It's possible to use firewalld everywhere, but then it might become difficult to manage because you would need to remove ufw first (and it might be already in use), etc...

I based myself on the table provided in the documentation (`TCP Port Configuration section` in `deployment-guide.md`) and in the issue #585 I opened a few days ago. The playbook `firewall.yml` will include the tasks from the file of the same name inside the `common` role. For each group, we define a list of ports that should be accessible and optionally a list of source from where the port needs to be reached. Without source, the firewall will be opened from any source (nginx). In case the source is the same IP as the current host on which the task is being executed, the firewall rule will not be set, as it means it's just a local communication (e.g. when the repo and search are on the same server, it won't put a rule for the Solr port).

I had to define new variables to contain the list of all hosts, even if at the moment the playbook only support 1 nginx for example, I prepare the variables to be able to have has many hosts as needed. Also, I couldn't use the already available variables anyway, because for me they are always null/undefined for remote hosts. I'm not sure if the issue is on my side or if it's a more generic topic, but from what I understand, `ansible_default_ipv4.address` is a fact-based variable, meaning that it's only available/defined for the server on which the task is currently running (where the facts were gathered). Therefore, I couldn't use that because all the variables for `repo_host`, `nginx_host`, etc. have a value of `127.0.0.1` if the host is a remote one. So for example, if a task is executed on `nginx`, then `nginx_host` would have its real IP, but `repo_host`, which should have a remote IP (another IP), actually has `127.0.0.1` because `ansible_default_ipv4.address` is undefined. So it wasn't working for the firewall setup since I needed to know all IPs from anywhere, to be able to run the `firewall.yml` playbook as a standalone. Therefore, I used `ansible_host`, as you can see inside the `roles/common/defaults/main.yml` file.

I added a parameter for the ADW port also, to avoid hardcoded references to `8880`. There are still a few hardcoded references inside the nginx for example, which I didn't touch so far, as there might be other things to take care of.

In regards to the Repository/Share clustering, do you think there is something specific to do for the firewall?

Any other comments/modifications/improvements that you could see in this? Don't hesitate to share and I can update my proposal if needed.

Thanks!

### DIFF
--- a/docs/deployment-guide.md
+++ b/docs/deployment-guide.md
@@ -263,12 +263,14 @@ Several roles setup services that listen on TCP ports and several roles wait for
 | repository                  | 8080        | nginx, search, syncservice                               | Yes                     |
 | search                      | 8983        | repository                                               | No                      |
 | transformers (aio t-engine) | 8090        | repository                                               | No                      |
+| transformers (router)       | 8095        | repository                                               | No                      |
+| transformers (sfs)          | 8099        | repository                                               | No                      |
 | syncservice                 | 9090        | nginx                                                    | No                      |
-| adw                         | 80          | nginx                                                    | No                      |
+| adw                         | 8880        | nginx                                                    | No                      |
 | nginx                       | 80          | `<client-ips>`                                           | No                      |
 | nginx                       | 443         | `<client-ips>`                                           | No                      |
 
-> NOTE: The transformers host will also contain the transform router process running on port 8095 and the shared file system process running on 8099 but communication between these components remains local.
+> NOTE: When using the ACS Community, some of these ports do not need to be opened (e.g. transform router/sfs, adw).
 
 ## Configure Your Deployment
 
@@ -750,10 +752,10 @@ To check your inventory file is configured correctly and the control node is abl
 ansible all -m ping -i inventory_ssh.yml
 ```
 
-**Optional** To check if the required ports for the deployment are available on the target machine and we also have connectivity between nodes (ex. repository connecting to the db on 5432) please run the prerequisite-checks playbook before you deploy ACS. If there are any firewalls blocking connectivity this playbook will discover them.
+**Optional** To check if the required ports for the deployment are available on the target machine and we also have connectivity between nodes (ex. repository connecting to the db on 5432) the prerun-network-checks playbook can be executed before you deploy ACS. If there are any firewalls blocking connectivity this playbook will discover them.
 
 ```bash
-pipenv run ansible-playbook playbooks/prerequisite-checks.yml -i inventory_ssh.yml
+pipenv run ansible-playbook playbooks/prerun-network-checks.yml -i inventory_ssh.yml [-e "@community-extra-vars.yml"]
 ```
 
 To deploy ACS 7 Enterprise on the target hosts execute the playbook as the current user using the following command:
--- a/playbooks/acs.yml
+++ b/playbooks/acs.yml
@@ -32,6 +32,9 @@
     - name: Setup
       ansible.builtin.setup:
 
+- name: Make sure the firewall is setup properly
+  ansible.builtin.import_playbook: firewall.yml
+
 - name: Database Role
   hosts: database
   gather_facts: no
--- a/playbooks/firewall.yml
+++ b/playbooks/firewall.yml
@@ -1,0 +1,112 @@
+---
+- name: Configure and enable the firewall on activemq
+  hosts: activemq
+  become: true
+  tasks:
+  - include_tasks: "../roles/common/tasks/firewall.yml"
+    vars:
+      firewall_ports:
+        - "{{ ports_cfg.activemq[activemq_protocol] }}"
+      firewall_sources:
+        - "{{ repo_hosts_list }}"
+        - "{{ sync_hosts_list }}"
+        - "{{ trans_hosts_list }}"
+        - "{{ search_ent_hosts_list }}"
+
+- name: Configure and enable the firewall on database
+  hosts: database
+  become: true
+  tasks:
+  - include_tasks: "../roles/common/tasks/firewall.yml"
+    vars:
+      firewall_ports:
+        - "{{ ports_cfg.postgres.sql }}"
+      firewall_sources:
+        - "{{ repo_hosts_list }}"
+        - "{{ sync_hosts_list }}"
+    when: repo_db_url == ""
+
+- name: Configure and enable the firewall on repository
+  hosts: repository
+  become: true
+  tasks:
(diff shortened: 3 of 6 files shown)


morgan-patou:
> --> Maybe I'm missing something but I don't see the variables using `ansible_default_ipv4.address` as working. So if you have the same issue, maybe these variables definition should be changed? In the `prerun-network-checks.yml` (formerly `prerequisite-checks.yml`) playbook, you are also using `ansible_host` and not the variables from the common role. So maybe it would be good to align all variables, to always use `ansible_host` (or `ansible_default_ipv4.address`), instead of sometimes one and sometimes the other?

Ok I just realized that in the [ACS playbook line 28](https://github.com/Alfresco/alfresco-ansible-deployment/blob/master/playbooks/acs.yml#L28), you are gathering the facts about all hosts first, and therefore the IPs of all hosts are now known. For my knowledge, isn't the line 28 actually not needed, since you are already gathering the facts at the [line 2 of the same ACS playbook](https://github.com/Alfresco/alfresco-ansible-deployment/blob/master/playbooks/acs.yml#L2)?

Would you prefer to use `ansible_default_ipv4.address` everywhere? If so, I could update the needed playbooks to not use `ansible_host` anymore but instead gather the facts for all hosts where needed. (`ansible_host` could potentially contain a hostname if I'm not wrong, so I assume it would be safer to use `ansible_default_ipv4.address`).

alxgomz:
Hi @morgan-patou ,

Adding firewall management rules was actually proposed and debated long ago and back, then PM. decided to not fiddle with that as there are so many different ways to setup firewalling rules to comply to users security approach.
The team changes deeply since this decision was made and I'm happy to re-trigger that conversation:
@gionn , @slohe1 , @vprovaggi , any insight?

morgan-patou:
> I like the idea of supporting a feature which increase the security of Alfresco installations managed by this playbook, however since the widespread usage of cloud infrastructures nowadays this feature would be beneficial only for installations in non-cloud environments (or at least for me, configuring firewall at operating system level when you must anyway configure the virtual networks it's just doing double work for nothing).

Depending on the cloud environment and specific image/OS chosen, the firewall might be enabled by default on the OS and blocking any external access, so not doing anything isn't really a solution either from my point of view, even on Cloud infra. Therefore, I believe that this would/could apply to both cloud and non-cloud usage of the playbook. Disabling the firewall completely isn't possible of course, therefore it's either not doing anything (providing the playbook but not executing it by default) or always setting up the firewall (and as you mentioned, on some clouds like AWS, this would/could be a double-work since the firewall/security will/might also be set on the cloud layer as well - but you have no insurance that the customers will setup their AWS properly...).


> Given this is not an easy topic, I would ask also to raise a different PR for the docs fixups which can for sure be handled faster than this one.

Will do

gionn:
> Disabling the firewall completely isn't possible of course, therefore it's either not doing anything (providing the playbook but not executing it by default) or always setting up the firewall

I would follow the approach of not doing anything firewall-related by default if it is not impacting the provisioning with an OS default configuration - which can be simplified to just configure the firewall on demand.

The second matter for this topic is: how do we test it.
I think the only choice we have now is to enable it on the [ec2 multimachine](https://github.com/Alfresco/alfresco-ansible-deployment/tree/master/molecule/multimachine) scenario because all the other ones are either running in docker or only single machine.

morgan-patou:
> I would follow the approach of not doing anything firewall-related by default if it is not impacting the provisioning with an OS default configuration - which can be simplified to just configure the firewall on demand and we will figure it later which are the OS that would benefit to have firewall configuration enabled by default.

As discussed previously, I created the PR#607 and moved the global doc update and the network checks part that already existed before (simple improvements for community for example & alignment with the doc as some parts were missing) into this new PR.

For this firewall topic, I added a flag `skip_fw_config` so that the firewall specific playbook isn't executed by default, unless someone pass it as parameter. I updated the doc as well to explain how to execute that before, during or after Alfresco deployment.


> The second matter for this topic is: how do we test it. I think the only choice we have now is to enable it on the [ec2 multimachine](https://github.com/Alfresco/alfresco-ansible-deployment/tree/master/molecule/multimachine) scenario because all the other ones are either running in docker or only single machine.

On my side, I tested the firewall configuration using our YaK on AWS and OCI (I didn't test on Azure but I could if needed) by deploying an environment composed of RedHat 8.6, Rocky Linux 8.6 and Ubuntu 22.04, all of them in the same inventory. I also tried to change where the Alfresco components were and the firewall playbook could be re-executed and reflected the changes properly (e.g. the repo host was initially on RHEL 8.6 so change the inventory to put the Ubuntu 22.04 as repo instead and re-launched the firewall playbook only).


Note: Currently the playbook do not remove existing firewall rules, it just makes sure that what is needed at the moment of execution is present. This is because we aren't really supposed to move components around once they are installed. In addition, it could be dangerous to do a cleanup since the firewall might contain rules coming from the customer directly. This also means that in case one of the IP changes for a specific host, then the playbook needs to be re-executed and the old IP would still appear in the rules I assume...

morgan-patou:
Some time ago, I noticed that Ansible wouldn't gather all the facts for `ansible_default_ipv4` even if `gather_facts: true` was used and that was causing the issue I was seeing before and the reason why I was using `ansible_host`... To keep consistency and since it's better to use `ansible_default_ipv4` rather than `ansible_host`, I put a small update to force ansible to gather the facts for all hosts in the inventory at the very beginning of the `firewall.yml` playbook with:
```yml
- name: Gather facts from all hosts
  hosts: all:!external
  gather_facts: false
  tasks:
    - name: Setup
      ansible.builtin.setup:
```

With that small update, the firewall configuration can be done more easily and it keeps the same nomenclature as what you did on your side as well for the host lists. There could be a small work to merge all host_lists to make it more pretty but I didn't want to touch your part for now, since this PR might not get approved anyway.

alxgomz:
Hi Morgan,

I took a very quick look at your PR (for the first time) and want to share some feedback.
I see you're leveraging the `common` role. This role is something I would now consider a relic from the past and I would avoid adding more stuff into it. The current approach would be to instead do all this computation at the playbook level (e.g. in included/imported tasks).
Another thing is that, if I try to think about this feature in terms of architecture, I would argue that ideally, the deifnition of network flows should be part of each role (e.g. solr role should mention its target port is 8983 and so on). I realize we're far from this point even without your PR! I just to emphasis how much such a PR will rely on an architecture we know needs to be refactored.

PR #590, opened 2023-05-15
OPSEXP-2112: add version checks for ATS mTLS  config

alxgomz:
skip mtls if versions do not support it


slohe1 (commented):
Can we try to fix this 
https://github.com/Alfresco/alfresco-ansible-deployment/actions/runs/4982770779/jobs/8925564908?pr=590#step:7:556

slohe1 (approved):

PR #594, opened 2023-05-18
Make opensearch suite create shorter resource name

gionn:
with the current max length limitation, opensearch resource on aws during IT get created as:

```
molecule-opensearch-opsexp-N
```

leaving just one digit to avoid collisions

### DIFF
--- a/molecule/opensearch/cleanup.yml
+++ b/molecule/opensearch/cleanup.yml
@@ -10,9 +10,12 @@
       {{ lookup('ansible.builtin.env', 'BRANCH_NAME') }}
     build: >-
       {{ lookup('ansible.builtin.env', 'BUILD_NUMBER') }}
-    ec2_resource_name: "{{ ['molecule', it_platform, repo_branch, build] | join('_') }}"
     domain_name: >-
-      {{ ['molecule', it_platform, build, repo_branch] | join('-') | lower | truncate(28, True, '') }}
+      {{ ['molecule', it_platform, build, repo_branch]
+      | join('-')
+      | ansible.builtin.regex_replace('[^a-zA-Z0-9]', build[-1])
+      | lower
+      | truncate(28, True, '') }}
   tasks:
     - name: Destroy Opensearch domain {{ domain_name }}
       community.aws.opensearch:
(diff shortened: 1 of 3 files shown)


molecule/opensearch/cleanup.yml
```diff
@@ -10,9 +10,9 @@
       {{ lookup('ansible.builtin.env', 'BRANCH_NAME') }}
     build: >-
       {{ lookup('ansible.builtin.env', 'BUILD_NUMBER') }}
-    aws_base_resource_name: "{{ ['molecule', it_platform, repo_branch, build] | join('_') }}"
+    ec2_resource_name: "{{ ['molecule', it_platform, repo_branch, build] | join('_') }}"
     domain_name: >-
-      {{ aws_base_resource_name | ansible.builtin.regex_replace('[^a-zA-Z0-9]','-') | lower | truncate(28, True, '') }}
+      {{ ['molecule', it_platform, build, repo_branch] | join('-') | lower | truncate(28, True, '') }}
```
gionn, line 15: not many alternative ideas:

I've swapped build and repo branch vars so at least we avoid truncating the build number which is the most variable thing in the resource naming, so instead of:

```
molecule_opensearch_opsexp_1
```
we get:

```
molecule_opensearch_123456_o
```

alxgomz: can't we deal we "transitions" problem manually to move from one naming model to another and replace the `molecule`prefix with a tag?

alxgomz (approved):

molecule/opensearch/cleanup.yml
```diff
@@ -10,9 +10,12 @@
       {{ lookup('ansible.builtin.env', 'BRANCH_NAME') }}
     build: >-
       {{ lookup('ansible.builtin.env', 'BUILD_NUMBER') }}
-    aws_base_resource_name: "{{ ['molecule', it_platform, repo_branch, build] | join('_') }}"
     domain_name: >-
-      {{ aws_base_resource_name | ansible.builtin.regex_replace('[^a-zA-Z0-9]','-') | lower | truncate(28, True, '') }}
+      {{ ['molecule', it_platform, build, repo_branch]
+      | join('-')
+      | ansible.builtin.regex_replace('[^a-zA-Z0-9]', build[-1])
```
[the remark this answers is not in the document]
alxgomz: last number of the build number (which seemed more random than the first one, and safer than "-" in case it happens to be the last character after trucantion)

Issue #438, opened 2022-09-24
Make the `helper_modules` role a simple task file


alxgomz:
We have introduced a while ago a playbook which can be run before running the deployment playbook, in order to check required TCP ports are available from and to the appropriate machines.
This has been implemented within a role. Which does the following:
* checks on the destination host (where the port needs to be available) that no other service is running
* starts listening on the port
* try connecting from the source to the opened port

While the logic is fine we think it doesn't make sense to use a role for that. The purpose of this ticket is to achieve:
* Removal of the role
* Move the custom `listen_port` module to the main playbook
* Creation a task file and call it from the `prerequisite-checks.yml` playbook

alxgomz:
tracked by OPSEXP-2147

PR #607, opened 2023-06-01
Update deployment guide components and network checks

morgan-patou:
In relation to the previous PR #589, I split the doc & network checks related improvements/corrections in this new PR.

Included in this PR:
- correction of the documentation and alignment between the doc and the network checks playbook that was available before but not doing all the checks
- addition of some conditions for execution of tasks in the network checks playbook in case it's being executed on Community
- Inclusion of the Alfresco Control Center on the documentation since I saw you pushed it recently into the master branch (c.f. #564)
- Inclusion of the Alfresco Control Center on the network checks playbook

Not included in this PR:
- Update of the `docs/resources/*.png` files. I believe that you generate the png files based on the puml. I suggested in this PR an update of the puml files, but I didn't re-generate the png, as I assume you have what is needed for that while I don't

Feel free to let me know if I need to update/remove something.


gionn (dismissed):
LGTM

gionn:
Fix for this failing e2e is already in master

```
  TASK [Compare host OS with supported matrix] ***********************************
  fatal: [search-enteprise-integration-instance]: FAILED! => {"changed": false, "msg": ["Rocky 8.8 is not a supported OS"]}
```

playbooks/prerun-network-checks.yml
```diff
@@ -1,133 +1,188 @@
...
       checked_port: "{{ ports_cfg.activemq[activemq_protocol] }}"
       delegate_target: "{{ groups.activemq | first }}"
+    when: groups.activemq | default([]) | length > 0
 
-- name: prerequisite checks
+- name: Run preliminary network checks for acc hosts
+  hosts: acc
+  become: yes
+  roles:
+    - role: '../roles/helper_modules'
```
alxgomz, line 125: Why are you calling the role if you include the tasks explicitly?
To be fair, that stuff should not even be a role. If you feel like moving it to the playbook level feel free to do so.
If not, then just remove the `roles` execution

alxgomz (changes requested):
Slight change

playbooks/prerun-network-checks.yml
morgan-patou: All the other play of this playbook do that, I simply duplicated the ADW play and changed all `adw` references with `acc`, since that seems to be what was required for ACC as well. Isn't that to load the library `listen_port.py` that is then used inside the `check_port.yml` task? If that isn't the reason, I do not know, do did it like that for other plays so I just did the same.

alxgomz: Oh lord yes you're most likely right. That's a really ugly hack :-/ 
I've done some quick tests and it seems a single role is needed for the complete set of plays in this playbook to find the module file.
That said, I have created a new ticket to completely get rid of that roles: OPSEXP-2147 (shadows #438)
For now we can move forward with current code in this PR.

alxgomz (approved):

PR #611, opened 2023-06-06
OPSEXP-2135 Add plantuml regeneration workflow

gionn:
Ref: OPSEXP-2135

Tested on https://github.com/Alfresco/alfresco-ansible-deployment/pull/610

### DIFF
--- a/.github/workflows/plantuml.yml
+++ b/.github/workflows/plantuml.yml
@@ -1,0 +1,39 @@
+name: Generate PlantUML Diagrams
+
+on:
+  pull_request:
+    paths:
+      - "**.puml"
+  push:
+    paths:
+      - "**.puml"
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/plantuml.yml
```diff
@@ -0,0 +1,39 @@
...
+
+on:
+  pull_request:
+    paths:
+      - "**.puml"
+  push:
+    paths:
+      - "**.puml"
+    branches:
+      - master
```
alxgomz, line 11: Do we really want to trigger on master push. diagram should anyway be updated in the PR before being merged and we should not (auto-commit) to master directly?

gionn: for the same reason you generally want to run tests again on the default branch after they already run on a PR - if multiple PRs get worked in parallel and get merged as they are ready it's always best to also run in master to avoid inconsistencies after merging the N+1 PR

gionn: and of course this workflow would just run when a commit contains puml files changes and it's pretty fast to run

alxgomz (approved):

PR #614, opened 2023-06-13
OPSEXP-2155 Make plantuml workflow autocommit configured as per docs

gionn:
As per https://github.com/stefanzweifel/git-auto-commit-action#use-in-forks-from-public-repositories

I tried directly on https://github.com/Alfresco/alfresco-ansible-deployment/pull/607 but it doesn't trigger when using `pull_request_target` and I suspect we need this merged on default branch before it get triggered

Ref: OPSEXP-2155

### DIFF
--- a/.github/workflows/plantuml.yml
+++ b/.github/workflows/plantuml.yml
@@ -1,7 +1,7 @@
 name: Generate PlantUML Diagrams
 
 on:
-  pull_request:
+  pull_request_target:
     paths:
       - "**.puml"
   push:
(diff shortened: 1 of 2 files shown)


.github/workflows/plantuml.yml
```diff
@@ -20,6 +20,9 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
+          # https://github.com/stefanzweifel/git-auto-commit-action#workflow-should-run-in-base-repository
```
alxgomz, line 23: This anchor point to nowhere

alxgomz (approved):

.github/workflows/plantuml.yml
gionn: that anchor is inside an accordion which is collapsed by default, will link to the parent 🤦🏻

PR #625, opened 2023-06-20
OPSEXP-1280: refactor acs repo properties file generation

alxgomz:
Ref: OPSEXP-1280

2 ideas behind this refactoring: 
* Using a "master" template to generate properties computed/computable from the architecture/playbook on one side and another one to add properties from user provided files . Finally use `assemble`  to generate the final config file
* Moving as many properties as possible to a yaml datastructure in the repository group so that we can:
  * simplify the properties template to just render key/values
  * move some inventory dependent variables outside of the repository role

### DIFF
--- a/docs/deployment-guide.md
+++ b/docs/deployment-guide.md
@@ -816,29 +816,40 @@ There are some useful argument you can use with `ansible-playbook` command in ma
 
 ## ACS cluster
 
-Due to load or high availability needs, you might want to deploy a cluster of several repository nodes. This can be achieved rather simply by:
+Due to load or high availability needs, you might want to deploy a cluster of
+several repository nodes. This can be achieved rather simply by:
 
-* Giving the playbook the location of the shared storage used for the ACS contentstore (See [Shared storage documentation](shared-contentstore.md) for details).
+* Giving the playbook the location of the shared storage used for the ACS
+  contentstore (See [Shared storage documentation](shared-contentstore.md) for
+  details).
 * Specifying several hosts within the repository hosts group
 
-> :warning: as mention in the [Alfresco official documentation](https://docs.alfresco.com/content-services/latest/admin/cluster/#scenarioredundancycluster), "All the servers in a cluster should have static IP addresses assigned to them".
-> Not meeting this pre-requisite won't prevent the playbook from working but the cluster might will most likely stop working in case one of the server in the architecture changes IP address.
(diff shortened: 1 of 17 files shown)


roles/repository/tasks/main.yml
```diff
@@ -371,22 +396,6 @@
...
-        Insert or Update custom properties added through
-        configuration_files/alfresco-global.properties
-      ansible.builtin.blockinfile:
-        path: "{{ settings_folder }}/classpath/alfresco-global.properties"
-        marker: "{mark} - ANSIBLE MANAGED BLOCK"
-        marker_begin: "### Begin - Custom user properties"
-        marker_end: "### End - Custom user properties"
-        insertafter: EOF
-        block: >-
-          {{ lookup('file', role_path + '/../../configuration_files/alfresco-global.properties') }}
```
gionn, line 384: update https://github.com/Alfresco/alfresco-ansible-deployment/blob/master/docs/playbook-upgrade.md that we are getting rid of this file and the properties should be moved to new structure

gionn (changes requested):

docs/deployment-guide.md
```diff
@@ -410,9 +410,19 @@ And then edit `vars/secrets.yml` to fill all the required arguments for the plug
...
+
+> This approach is now discoraged and you should prefer using the [`repository`
+> group vars](../configuration_files/alfresco-global.properties)
+> `global_properties` as much as possible otherwise reference you own snippets
+> of properties file using either the new `repository` group var
+> `properties_snippets` or directly the `repository` role argument
+> `raw_properties`.
+
+`alfresco-global.properties` will be located in
+`/etc/opt/alfresco/content-services/classpath`.
```
gionn:
```suggestion
You can provide your [repository
configuration](https://github.com/Alfresco/acs-deployment/blob/master/docs/properties-reference.md)
by editing the `configuration_files/alfresco-global.properties` file.

> This approach is now discouraged and you should prefer using the [`repository`
> group vars](../configuration_files/alfresco-global.properties)
> `global_properties` as much as possible otherwise reference you own snippets
> of properties file using either the new `repository` group var
> `properties_snippets` or directly the `repository` role argument
> `raw_properties`.

`alfresco-global.properties` will be located in
`/etc/opt/alfresco/content-services/classpath`.
```

gionn (approved):

PR #638, opened 2023-07-12
Wait for mounts before starting components + add auto-restart on failure

morgan-patou:
As a best practice, the 4 standard paths mentioned [in the doc](https://github.com/Alfresco/alfresco-ansible-deployment/blob/master/docs/deployment-guide.md#folder-structure) will most probably be using a dedicated mount/filesystem.

At installation, it is fine since the filesystems should already be mounted (if some needs to be used). However, in case of OS restart, the different services are currently configured to be executed `After=syslog.socket network.target`. With this configuration of the service, it can happen that the OS tries to start the Alfresco components before the filesystem is even available. In such cases, it would obviously fail (`status=203/EXEC`) and therefore not startup the component at all. Example of logs showing that the filesystem for /opt/alfresco/ is mounted after ActiveMQ tries to startup:
```bash
[root@mop-alf-ce-tool ~]# grep After /etc/systemd/system/activemq.service
After=syslog.socket network.target
[root@mop-alf-ce-tool ~]#
[root@mop-alf-ce-tool ~]# journalctl -xe | grep -E "activemq|opt-alfresco"
-- Subject: Unit activemq.service has begun start-up
-- Unit activemq.service has begun starting up.
Jul 12 09:23:41 mop-alf-ce-tool systemd[904]: activemq.service: Failed to execute command: No such file or directory
Jul 12 09:23:41 mop-alf-ce-tool systemd[904]: activemq.service: Failed at step EXEC spawning /opt/alfresco/activemq.sh: No such file or directory
-- Subject: Process /opt/alfresco/activemq.sh could not be executed
-- The process /opt/alfresco/activemq.sh could not be executed and failed.
Jul 12 09:23:41 mop-alf-ce-tool systemd[1]: activemq.service: Control process exited, code=exited status=203
Jul 12 09:23:41 mop-alf-ce-tool systemd[1]: activemq.service: Failed with result 'exit-code'.
-- The unit activemq.service has entered the 'failed' state with result 'exit-code'.
-- Subject: Unit activemq.service has failed
-- Unit activemq.service has failed.
-- Subject: Unit opt-alfresco.mount has begun start-up
-- Unit opt-alfresco.mount has begun starting up.
[root@mop-alf-ce-tool ~]#
```

Since the filesystem is mounted after ActiveMQ tries to start, it fails with `No such file or directory` and because there is no restart by default, then ActiveMQ isn't running. I faced this issue on both `activemq.service` as well as `alfresco-content.service` but it can happen on most of them, depending on where they are installed and the order the OS decides for the startup.

Therefore, I believe it would be better to make sure that the local and remote filesystems are all properly mounted before trying to startup the different Alfresco components. This can be achieved by adding `local-fs.target remote-fs.target` in the definition of the different services (`After=syslog.socket network.target local-fs.target remote-fs.target`).

As part of this PR, I also added the automatic restart of the processes `on-failure` (c.f. [definition of the systemd service](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Restart=)). This is more of a debatable option, since the OS will try to restart most of the components even if you are shutting them down manually (without using the `systemctl stop xxx` command) on purpose because the `SuccessExitStatus` is usually not 0 (and I didn't configure it). This means that people should only use the service to stop/start the different components (which, from my point of view, should always be how it's done...). It adds recovery in case of OOM or other kills, which is a necessary thing from my point of view.

Please let me know if you would like to change slightly the configuration and/or remove the restart option I added.

### DIFF
--- a/roles/sfs/templates/alfresco-shared-fs.service
+++ b/roles/sfs/templates/alfresco-shared-fs.service
@@ -1,12 +1,14 @@
 [Unit]
 Description=Alfresco Shared File Store
-After=syslog.socket network.target
+After=syslog.socket network.target local-fs.target remote-fs.target
 
 [Service]
 Type=simple
 User={{ username }}
 ExecStart={{ binaries_folder }}/ats-shared-fs.sh
 WorkingDirectory={{ ats_home }}
+Restart=on-failure
+RestartSec=60
 
 [Install]
 WantedBy=multi-user.target
--- a/roles/transformers/templates/alfresco-tengine-aio.service
+++ b/roles/transformers/templates/alfresco-tengine-aio.service
@@ -1,12 +1,14 @@
 [Unit]
 Description=Alfresco Transform Service - AIO Transform Engine
-After=syslog.socket network.target
+After=syslog.socket network.target local-fs.target remote-fs.target
 
 [Service]
 Type=simple
 User={{ username }}
 ExecStart={{ binaries_folder }}/ats-ate-aio.sh
 WorkingDirectory={{ ats_home }}
+Restart=on-failure
+RestartSec=60
 
 [Install]
 WantedBy=multi-user.target
--- a/roles/activemq/templates/activemq.service.j2
+++ b/roles/activemq/templates/activemq.service.j2
@@ -1,12 +1,14 @@
 [Unit]
 Description=Apache ActiveMQ - Alfresco instance
-After=syslog.socket network.target
+After=syslog.socket network.target local-fs.target remote-fs.target
 
 [Service]
 Type=forking
 User={{ username }}
 ExecStart={{ binaries_folder }}/activemq.sh start
 ExecStop={{ binaries_folder }}/activemq.sh stop
+Restart=on-failure
+RestartSec=60
 
 [Install]
 WantedBy=multi-user.target
--- a/roles/search/templates/alfresco-search.service
+++ b/roles/search/templates/alfresco-search.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=Alfresco Search Service
-After=syslog.target network.target remote-fs.target nss-lookup.target
+After=syslog.target network.target local-fs.target remote-fs.target nss-lookup.target
 
 [Service]
 Type=forking
@@ -12,6 +12,7 @@ ExecStart={{ binaries_dir }}/solr.sh start -a "-Dcreate.alfresco.defaults={{ sea
 ExecReload={{ binaries_dir }}/solr.sh restart
 ExecStop={{ binaries_dir }}/solr.sh stop
 Restart=on-failure
+RestartSec=60
 
 [Install]
 WantedBy=multi-user.target
--- a/roles/search_enterprise/templates/elasticsearch-connector-reindex.service.j2
+++ b/roles/search_enterprise/templates/elasticsearch-connector-reindex.service.j2
@@ -1,6 +1,6 @@
 [Unit]
 Description=Alfresco Enterprise Search - Reindex job
-After=syslog.target network.target remote-fs.target nss-lookup.target
+After=syslog.target network.target local-fs.target remote-fs.target nss-lookup.target
 
 [Service]
 Type=oneshot
--- a/roles/search_enterprise/templates/elasticsearch-connector.service.j2
+++ b/roles/search_enterprise/templates/elasticsearch-connector.service.j2
@@ -1,6 +1,6 @@
 [Unit]
 Description=Alfresco Enterprise Search - All in one Service
-After=syslog.target network.target remote-fs.target nss-lookup.target
+After=syslog.target network.target local-fs.target remote-fs.target nss-lookup.target
 
 [Service]
 Type=simple
--- a/roles/sync/templates/alfresco-sync.service
+++ b/roles/sync/templates/alfresco-sync.service
@@ -1,12 +1,14 @@
 [Unit]
 Description=Alfresco Sync Service
-After=syslog.socket network.target
+After=syslog.socket network.target local-fs.target remote-fs.target
 
 [Service]
 Type=forking
 User={{ username }}
 ExecStart={{ binaries_folder }}/syncservice.sh start
 ExecStop={{ binaries_folder }}/syncservice.sh stop
+Restart=on-failure
+RestartSec=60
 
 [Install]
 WantedBy=multi-user.target
--- a/roles/tomcat/templates/alfresco-content.service
+++ b/roles/tomcat/templates/alfresco-content.service
@@ -1,12 +1,14 @@
 [Unit]
 Description=Alfresco Content Services
-After=syslog.socket network.target
+After=syslog.socket network.target local-fs.target remote-fs.target
 
 [Service]
 Type=forking
 User={{ username }}
 ExecStart={{ binaries_folder }}/tomcat.sh start
 ExecStop={{ binaries_folder }}/tomcat.sh stop
+Restart=on-failure
+RestartSec=60
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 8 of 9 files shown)


roles/sfs/templates/alfresco-shared-fs.service
```diff
@@ -1,12 +1,14 @@
...
-After=syslog.socket network.target
+After=syslog.socket network.target local-fs.target remote-fs.target
 
 [Service]
 Type=simple
 User={{ username }}
 ExecStart={{ binaries_folder }}/ats-shared-fs.sh
 WorkingDirectory={{ ats_home }}
+Restart=on-failure
+RestartSec=60
```
gionn, line 11: like alfresco-tengine-aio.service, but we probably want to add an `exec` in `ats-shared-fs.sh` before the java command to replace the previous bash with java process

gionn (changes requested):
see comments related to services which main process is a java process and its exit status behaviour when receiving a sigterm from systemd https://serverfault.com/a/695863/72778

overall, waiting for fs is indeed a must have

morgan-patou:
Yes java will exit 143, but if we add the `SuccessExitStatus` configuration, we might miss some non-expected termination of the processes. If I'm not wrong, an OS OOM kill might first send a SIGTERM to the Java process before the SIGKILL. Therefore I would assume it to be possible for Java to exit 143 even if it was killed by the OS OOM Killer. And in such case, we would want the process to automatically restart. I might be wrong, but that's basically why I didn't set the `SuccessExitStatus` one.


With the proposed PR as of now:
- ActiveMQ manual shutdown
```bash
[root@mop-alf-ce-tool ~]# systemctl status activemq
● activemq.service - Apache ActiveMQ - Alfresco instance
   Loaded: loaded (/etc/systemd/system/activemq.service; enabled; vendor preset: disabled)
   Active: active (running) since Wed 2023-07-12 11:49:33 CEST; 3h 32min ago
  Process: 1089 ExecStart=/opt/alfresco/activemq.sh start (code=exited, status=0/SUCCESS)
 Main PID: 1198 (java)
[root@mop-alf-ce-tool ~]# 
[alfresco@mop-alf-ce-tool ~]$ /opt/alfresco/activemq.sh stop
...
Connecting to pid: 1198
.Stopping broker: localhost
.. FINISHED
[alfresco@mop-alf-ce-tool ~]$
[root@mop-alf-ce-tool ~]# systemctl status activemq
● activemq.service - Apache ActiveMQ - Alfresco instance
   Loaded: loaded (/etc/systemd/system/activemq.service; enabled; vendor preset: disabled)
   Active: activating (auto-restart) (Result: exit-code) since Wed 2023-07-12 15:22:06 CEST; 7s ago
  Process: 7013 ExecStop=/opt/alfresco/activemq.sh stop (code=exited, status=1/FAILURE)
  Process: 1089 ExecStart=/opt/alfresco/activemq.sh start (code=exited, status=0/SUCCESS)
 Main PID: 1198 (code=exited, status=0/SUCCESS)
```

- tengine-aio manual kill
```bash
[root@mop-alf-ce-tool ~]# systemctl status alfresco-tengine-aio
● alfresco-tengine-aio.service - Alfresco Transform Service - AIO Transform Engine
   Loaded: loaded (/etc/systemd/system/alfresco-tengine-aio.service; enabled; vendor preset: disabled)
   Active: active (running) since Wed 2023-07-12 11:49:31 CEST; 3h 36min ago
 Main PID: 1088 (java)
[root@mop-alf-ce-tool ~]#
[root@mop-alf-ce-tool ~]# kill 1088
[root@mop-alf-ce-tool ~]# 
[root@mop-alf-ce-tool ~]# systemctl status alfresco-tengine-aio
● alfresco-tengine-aio.service - Alfresco Transform Service - AIO Transform Engine
   Loaded: loaded (/etc/systemd/system/alfresco-tengine-aio.service; enabled; vendor preset: disabled)
   Active: activating (auto-restart) (Result: exit-code) since Wed 2023-07-12 15:27:33 CEST; 1s ago
  Process: 1088 ExecStart=/opt/alfresco/ats-ate-aio.sh (code=exited, status=143)
 Main PID: 1088 (code=exited, status=143)
```

- Solr (`alfresco-search`) also has a status of `1/Failure` if it is shutdown using `/opt/alfresco/search-services-2.0.6.1/solr.sh stop`

- ACS tomcat doesn't stop in time if you use the `/opt/alfresco/tomcat.sh stop` command. Because it's using the default parameters, the command will return too soon (timeout after 5s), saying that Tomcat wasn't stopped in time
```bash
[alfresco@mop-alf-ce-repo ~]$ /opt/alfresco/tomcat.sh stop
Using CATALINA_BASE:   /etc/opt/alfresco/tomcat
Using CATALINA_HOME:   /opt/apache-tomcat-9.0.59
...
Tomcat did not stop in time.
PID file was not removed.
To aid diagnostics a thread dump has been written to standard out.
[alfresco@mop-alf-ce-repo ~]$
```

Therefore the Java process will remain forever and never be removed for Tomcat and it's just not shutting down at all. Therefore the OS Service doesn't detect any failure either but Alfresco isn't reachable anymore.

So this script doesn't really appear to be working properly for me. On my side, I usually use `${CATALINA_HOME}/bin/startup.sh` to start Tomcat and then `${CATALINA_HOME}/bin/shutdown.sh 60 -force` to shut it down, I don't use the `catalina.sh` directly as you do inside `/opt/alfresco/tomcat.sh` but the wrappers that Tomcat provides.

alxgomz:
Reading at the [systemd units documentation](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Options) I would argue that using `simple` types of unit is actually plain wrong.
Sounds to me that we should instead be using `exec` or `forking` (and make sure wrapper scripts use the `exec` command.
Wdyt? Do you read that as I do?

Even beter I woud argue tat none of these wrapper script really are necessary as they mostly set environment vars, which a unit can easily do... So in the end the best approach may be to get rid of them, stick to `simple` unit types and have `ExecStart` really start the service instead of calling a wrapper shell.

Though I realise this is more work and can understand if that's not something you want to touch on @morgan-patou . Also it has implication with several issuies we have opened for a while regarding removing the infamous `setenv.sh`

alxgomz:
btw, @morgan-patou, that kind of feedback is invaluable to us! Thanks a lot!

roles/sfs/templates/alfresco-shared-fs.service
alxgomz: I think we should move away from simple, or get away from wrapper scripts instead

roles/transformers/templates/alfresco-tengine-aio.service
```diff
@@ -1,12 +1,14 @@
...
-After=syslog.socket network.target
+After=syslog.socket network.target local-fs.target remote-fs.target
 
 [Service]
 Type=simple
 User={{ username }}
 ExecStart={{ binaries_folder }}/ats-ate-aio.sh
 WorkingDirectory={{ ats_home }}
+Restart=on-failure
+RestartSec=60
```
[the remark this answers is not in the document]
alxgomz: `143` is the regular exit code triggered by SIGTERM and SIGTERM is already considered a successful outcome ( as per https://www.freedesktop.org/software/systemd/man/systemd.service.html#SuccessExitStatus=)
HEnce I don't think it's necessary to add it

alxgomz (changes requested):
For now the quick win sounds like using `exec`unit types

alxgomz (approved):
Approving as it fixes the reported issue and further changees to the uint types is actually a separate enhancement

alxgomz:
enterprise checks: https://github.com/Alfresco/alfresco-ansible-deployment/actions/runs/5533601854

gionn (approved):

gionn:
Thanks for this contribution as well!

PR #642, opened 2023-07-17
OPSEXP-2203 Restart affected services by certificate refresh for mtls

gionn:
Ref: OPSEXP-2203

### DIFF
--- a/roles/java/tasks/keystores.yml
+++ b/roles/java/tasks/keystores.yml
@@ -165,3 +165,6 @@
           -alias {{ item.name }}
       tags:
         - molecule-idempotence-notest
+
+- name: Force all notified handlers to run
+  ansible.builtin.meta: flush_handlers
--- a/roles/java/tasks/keystores.yml
+++ b/roles/java/tasks/keystores.yml
@@ -28,7 +28,6 @@
         mode: "0400"
         dest: "{{ download_location }}"
       register: certcp
-      notify: restart-java-services
     - name: Export certificate chain
       loop: "{{ java_keystore.cert_containers }}"
       loop_control:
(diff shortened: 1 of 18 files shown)


roles/repository/tasks/restart.yml
```diff
@@ -0,0 +1,9 @@
+- name: Restart {{ svc }}
+  become: true
+  throttle: 1
+  vars:
+    svc: alfresco-content{% if groups.repository | length > 1 %}-monitored-startup{% endif %}
+  ansible.builtin.service:
+    name: "{{ svc }}"
+    state: restarted
+    daemon_reload: true
```
gionn, line 9: not really needed?

```suggestion
```

roles/java/tasks/keystores.yml
```diff
@@ -28,6 +28,7 @@
         mode: "0400"
         dest: "{{ download_location }}"
       register: certcp
+      notify: restart-java-services
```
alxgomz, line 31: I thikn this handler would be better placed in the task "Import certificates to a keystore" & "Trust certificates CA"

roles/java/tasks/keystores.yml
```diff
@@ -164,3 +165,6 @@
           -alias {{ item.name }}
       tags:
         - molecule-idempotence-notest
+
+- name: Force all notified handlers to run
+  ansible.builtin.meta: flush_handlers
```
alxgomz, line 170: Why do we need that? Ins't it OK to run at the end of the play?

roles/repository/tasks/restart.yml
alxgomz: should not indeed but better to keep it i case notrification order  changes or we one day need to include just that file

alxgomz (commented):
just wondering about the flush_handler usage

roles/java/tasks/keystores.yml
alxgomz: make sense indeed

alxgomz (approved):

PR #644, opened 2023-07-19
OPSEXP-2119 Activemq 5.17 support on playbook

slohe1:
OPSEXP-2119

### DIFF
--- a/roles/activemq/meta/argument_specs.yml
+++ b/roles/activemq/meta/argument_specs.yml
@@ -11,5 +11,5 @@ argument_specs:
           type: "str"
         activemq_log_config:
           description: activemq log4j and log4j2 config file.
-          default: {% if activemq_version >= '5.17.0' %}log4j2.properties{% else %}log4j.properties{% endif %}"
+          default: "{% if activemq_version >= '5.17.0' %}log4j2.properties{% else %}log4j.properties{% endif %}"
           type: "str"
--- a/roles/activemq/meta/argument_specs.yml
+++ b/roles/activemq/meta/argument_specs.yml
@@ -11,5 +11,5 @@ argument_specs:
           type: "str"
         activemq_log_config:
           description: activemq log4j and log4j2 config file.
-          default: "{{ 'log4j2.properties' if activemq_version >= '5.17.0' else 'log4j.properties' }}"
+          default: {% if activemq_version >= '5.17.0' %}log4j2.properties{% else %}log4j.properties{% endif %}"
           type: "str"
--- a/roles/activemq/meta/argument_specs.yml
+++ b/roles/activemq/meta/argument_specs.yml
@@ -6,10 +6,12 @@ argument_specs:
     options:
         activemq_version:
           description: activemq version to use for deployment.
-          default: "5.17.6"
           required: false
-          type: "str"
+          default: "5.17.4"
         activemq_log_config:
           description: activemq log4j and log4j2 config file.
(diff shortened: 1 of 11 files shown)


alxgomz (changes requested):
ActiveMQ 5.17 is only supported since acs 7.3.
We need a backward compatible approach

alxgomz:
I would propose to make the activemq versiojn a [role argument](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_reuse_roles.html#role-argument-validation) passed by `acs.yml` and switch log file based on that

roles/activemq/meta/argument_specs.yml
```diff
@@ -0,0 +1,14 @@
...
+    author: Alfresco
+    options:
+      activemq_version:
+        description: activemq version to use for deployment.
+        required: false
+        default: 5.17.4
+      activemq_log_config:
+        description: activemq log4j and log4j2 config file.
+        required: false
+        default: log4j2
```
alxgomz, line 14: This should be a role argument only if it allow you to configure activemq to use log4j or log4jv2 regardless of the version but that's not possible.
As this variable is not used anywhere else but in a single task would make it a task variable only, not a role argument.

alxgomz (changes requested):

roles/activemq/tasks/main.yml
```diff
@@ -3,7 +3,7 @@
   ansible.builtin.get_url:
     url: "{{ dependencies_url.activemq }}"
     checksum: sha512:{{ dependencies_url.activemq_sha512_checksum_url }}
-    dest: "{{ download_location }}/apache-activemq-{{ dependencies_version.activemq }}-bin.tar.gz"
+    dest: "{{ download_location }}/apache-activemq-{{ activemq_version }}-bin.tar.gz"
```
alxgomz, line 6: 👍

roles/activemq/tasks/main.yml
```diff
@@ -100,8 +100,10 @@
       no_log: true
 
     - name: Change location of ActiveMQ logs
+      vars:
+        activemq_log_config: "{{ 'log4j2' if activemq_version >= '5.17.0' else 'log4j' }}"
```
alxgomz, line 104: Too bad you're still not using the `version` filter mentioned by @gionn . You could end up with unexpected results.
Check the examples https://docs.ansible.com/ansible/latest/collections/ansible/builtin/version_test.html#examples

alxgomz (changes requested):

roles/activemq/tasks/main.yml
slohe1: True, I missed that one.

alxgomz (approved):

roles/activemq/tasks/main.yml
```diff
@@ -100,8 +100,10 @@
       no_log: true
 
     - name: Change location of ActiveMQ logs
+      vars:
+        activemq_log_config: "{% if activemq_version is version('5.17.0', '>=') %}log4j2{% else %}log4j{% endif %}"
       ansible.builtin.replace:
-        path: "{{ activemq_conf }}/log4j.properties"
+        path: "{{ activemq_conf }}/{{ activemq_log_config }}.properties"
         regexp: "activemq.data"
         replace: "activemq.log"
```
gionn, line 108: not sure what is this task exactly doing but ok 😬

gionn (changes requested):

gionn (approved):

PR #669, opened 2023-09-19
OPSEXP-2303 Add identity role for SSO

gionn:
Ref: OPSEXP-2303


roles/identity/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+# defaults file for identity
+identity_admin_username: admin
+identity_admin_password: alfrescoalfrescoadmin
```
alxgomz, line 4: can we make it a mandatory role arg?

roles/identity/tasks/realm.yml
```diff
@@ -0,0 +1,41 @@
+- name: Fetch Alfresco theme
+  become: true
+  vars:
+    alfresco_theme_version: 0.3.5
```
alxgomz, line 4: please make it a role arg

roles/identity/tasks/realm.yml
```diff
@@ -0,0 +1,41 @@
...
+  ansible.builtin.unarchive:
+    src: "https://github.com/Alfresco/alfresco-keycloak-theme/releases/download/
+      {{ alfresco_theme_version }}/alfresco-keycloak-theme-{{ alfresco_theme_version }}.zip"
+    dest: /opt/keycloak/keycloak-{{ identity_keycloak_quarkus_version }}/themes
+    remote_src: true
+
+- name: Configure Alfresco Keycloak realm
+  community.general.keycloak_realm:
+    auth_client_id: admin-cli
+    auth_keycloak_url: http://localhost:8080
```
alxgomz, line 14: Eventually that should be read from known_url but TBH I have no idea how this will all play together with the reverse proxy so good enough for I guess

alxgomz (commented):
some comment. Can be done in another PR if you prefer

roles/identity/tasks/realm.yml
gionn: all auth_* here are connection details to reach the keycloak api run from the identity host so localhost should be good to go (port needs to be handled better but at a later stage)

roles/identity/tasks/realm.yml
gionn: I'd prefer not since is not something that the average role user want to explicitly change

roles/identity/tasks/realm.yml
alxgomz: oh ok got it. I. didn't check the doc for these params tbh

alxgomz (approved):

PR #673, opened 2023-09-22
OPSEXP-2304 Configure SSO in repository

gionn:
Ref: OPSEXP-2304

### DIFF
--- a/group_vars/repository.yml
+++ b/group_vars/repository.yml
@@ -33,7 +33,7 @@ global_properties:
     cluster:
       enabled: "{{ (groups['repository'] | length > 1 and not (cluster_keepoff | bool)) | lower }}"
   share:
-    host: "{{ fqdn_alfresco | default(known_urls[0]) | default(nginx_host) }}"
+    host: "{{ fqdn_alfresco | default(known_urls[0] | urlsplit('hostname')) | default(nginx_host) }}"
     port: "{{ acs_play_port }}"
     protocol: "{{ acs_play_proto }}"
   messaging:
--- a/group_vars/repository.yml
+++ b/group_vars/repository.yml
@@ -27,7 +27,7 @@ global_properties:
     username: "{{ repo_db_username }}"
     password: "{{ repo_db_password }}"
   alfresco:
-    host: "{{ fqdn_alfresco | default(nginx_host) }}"
+    host: "{{ fqdn_alfresco | default(known_urls[0] | urlsplit('hostname')) | default(nginx_host) }}"
     port: "{{ acs_play_port }}"
     protocol: "{{ acs_play_proto }}"
     cluster:
--- a/molecule/identity/molecule.yml
+++ b/molecule/identity/molecule.yml
@@ -1,0 +1,48 @@
+---
+dependency:
+  name: galaxy
+driver:
+  name: docker
+platforms:
+  - name: identity-instance
+    image: $MOLECULE_ROLE_IMAGE
+    dockerfile: ../../tests/molecule/Dockerfile-noprivs.j2
+    command: "/lib/systemd/systemd"
+    privileged: true
+    tmpfs:
+      - /run
+      - /run/lock
+      - /tmp
+    volume_mounts:
+      - "/sys/fs/cgroup:/sys/fs/cgroup:ro"
+    groups:
+      - database
+      - activemq
+      - repository
+      - trusted_resource_consumers
+      - identity
+      - nginx
+    published_ports:
+      - 80/tcp
+      - 0.0.0.0:443:443/tcp
+      - 0.0.0.0:8080:8080/tcp
+
+provisioner:
+  name: ansible
+  config_options:
+    defaults:
+      pipelining: True
+  ansible_args:
+    - -e
+    - "@tests/test-extra-vars.yml"
+    - -e
+    - "@tests/test-ssl.yml"
+  inventory:
+    links:
+      group_vars: ../../group_vars
+      host_vars: host_vars
+  playbooks:
+    prepare: ../default/prepare.yml
+    verify: ../default/verify.yml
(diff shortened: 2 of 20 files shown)


molecule/identity/molecule.yml
```diff
@@ -0,0 +1,47 @@
...
+      - database
+      - activemq
+      - repository
+      - trusted_resource_consumers
+      - identity
+      - nginx
+    published_ports:
+      - 80/tcp
+      - 0.0.0.0:443:443/tcp
+      - 0.0.0.0:8080:8080/tcp
```
alxgomz, line 28: any reason why port 80 is mapped dynamically and not others?

molecule/identity/verify.yml
```diff
@@ -0,0 +1,50 @@
...
+          - ansible_facts.services['alfresco-content.service'].state == "running"
+          - ansible_facts.services['keycloak.service'].state == "running"
+
+    - name: Retrieve contents of alfresco-global.properties
+      become: true
+      ansible.builtin.slurp:
+        src: /etc/opt/alfresco/content-services/classpath/alfresco-global.properties
+      register: slurp_global_properties
+
+    - name: Check reindex service contains the expected ExecStart line
```
alxgomz, line 21:
```suggestion
    - name: Check repository service unit contains the expected ExecStart line
```

molecule/identity/verify.yml
```diff
@@ -0,0 +1,50 @@
...
+      vars:
+        global_properties_content: "{{ slurp_global_properties['content'] | b64decode }}"
+        expected_auth_chain: "authentication.chain=identity-service1:identity-service,alfrescoNtlm1:alfrescoNtlm"
+        expected_service_resource: "identity-service.resource=alfresco"
+        expected_service_credentials: "identity-service.credentials.secret="
+        expected_auth_url: "identity-service.auth-server-url=http://172.17.0.2:8082/"
```
alxgomz, line 27: That relies a lot on Docker default behvior. Maybe use a `regex_search` filter instead?

playbooks/acs.yml
```diff
@@ -176,11 +186,16 @@
       when:
         - repo_mtls_required
         - ats_mtls_capable
+    - name: Enable identity service
+      when: ((groups.identity | default([])) + (groups.external_identity | default([]))) | length > 0
+      ansible.builtin.set_fact:
+        identity_url: "http://{{ identity_host }}:{{ ports_cfg.identity.http }}/"
```
alxgomz, line 192: 🤞

roles/nginx/templates/alfresco_proxy.j2
```diff
@@ -71,6 +70,16 @@
...
+
+    location /auth/ {
+        proxy_pass http://{{ identity_host }}:{{ ports_cfg.identity.http }}/;
+        proxy_redirect off;
+        proxy_buffering off;
+        proxy_set_header Host              $host:$server_port;
+        proxy_set_header X-Real-IP         $remote_addr;
+        proxy_set_header X-Forwarded-For   $proxy_add_x_forwarded_for;
+        proxy_set_header X-Forwarded-Proto $scheme;
         proxy_pass_header Set-Cookie;
```
alxgomz, line 83: If I understood coorectly how edge mode works I think we should not pass this header at the nginx level. Those should be set by the AWS LB which endorse the SSL termintation

roles/repository/defaults/main.yml
```diff
@@ -84,3 +84,8 @@ keystore_src: >-
   {%- endif %}
 repo_keystore: {}
 repository_monitored_startup_timeout_seconds: 300
+
+# Identity service arguments
+repository_identity_url: ''
+repository_identity_client_id: alfresco
+repository_identity_client_secret: null
```
alxgomz, line 91: it's defiend as a str in the arg spec so an empty string would make more sense

roles/repository/tasks/main.yml
```diff
@@ -345,12 +345,31 @@
...
+      when: repository_identity_url | length > 0
+      vars:
+        sso_repository_properties:
+          authentication:
+            chain: identity-service1:identity-service,alfrescoNtlm1:alfrescoNtlm
+          identity-service:
+            resource: "{{ repository_identity_client_id }}"
+            credentials:
+              secret: "{{ repository_identity_client_secret }}"
+          # Do not try to merge with the previous `identity-service` or you will regret it
```
alxgomz, line 362: why?

alxgomz (changes requested):

molecule/identity/molecule.yml
gionn: mostly copy paste from similar suites

molecule/identity/verify.yml
gionn: oops

roles/nginx/templates/alfresco_proxy.j2
gionn: nginx config should be flexible enough to do the right when there is another reverse proxy in front of it, let me do some research because the current config is a plain copy-paste of what we already have here

group_vars/repository.yml
```diff
@@ -33,7 +33,7 @@ global_properties:
     cluster:
       enabled: "{{ (groups['repository'] | length > 1 and not (cluster_keepoff | bool)) | lower }}"
   share:
-    host: "{{ fqdn_alfresco | default(known_urls[0]) | default(nginx_host) }}"
+    host: "{{ fqdn_alfresco | default(known_urls[0] | urlsplit('hostname')) | default(nginx_host) }}"
```
gionn, line 36: check this

not sure if I should uniform to L30 (by adding know_urls there or removing here)

alxgomz (approved):

PR #674, opened 2023-09-26
OPSEXP-2304 Fixup community workflow

gionn:
Ref: OPSEXP-2304

### DIFF
--- a/roles/identity/tasks/realm.yml
+++ b/roles/identity/tasks/realm.yml
@@ -38,6 +38,8 @@
       - ru
       - sv
       - zh-CN
+  tags:
+    - molecule-idempotence-notest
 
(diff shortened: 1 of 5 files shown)


roles/identity/tasks/realm.yml
```diff
@@ -50,6 +49,4 @@
     realm: alfresco
     client_id: alfresco
     enabled: true
-    redirect_uris: "{{ known_urls | map('regex_replace', '(.*)$', '\\1*') | list }}"
-    web_origins: "{{ known_urls }}"
```
gionn, line 54: need to move them as role argument - will do in the other PR

alxgomz (approved):

PR #675, opened 2023-09-27
OPSEXP-2305 Refactor frontend apps in a new webapp role

gionn:
Ref: OPSEXP-2305


alxgomz (changes requested):
Exactly how I think we should be using roles! 👍 💅 
I'm just concerned about the role and its vars names :-/

alxgomz (approved):

PR #677, opened 2023-10-02
OPSEXP-2305 Enable SSO for adf apps

gionn:
Ref: OPSEXP-2305

### DIFF
--- a/molecule/identity/verify.yml
+++ b/molecule/identity/verify.yml
@@ -2,6 +2,10 @@
 - name: Verify
   hosts: all
   gather_facts: false
+  vars:
+    adf_apps:
+      - alfresco-digital-workspace
+      - alfresco-control-center
   tasks:
     - name: Populate services facts
       ansible.builtin.service_facts:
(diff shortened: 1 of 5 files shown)


molecule/identity/verify.yml
alxgomz, line 1: DTAS seems to have identity related scenarios. We should add them here:
https://github.com/Alfresco/alfresco-deployment-test-automation-scripts/blob/master/tests/scenarios/test_identity.py
If more work is needed then we would need a dedicated task

playbooks/acs.yml
```diff
@@ -165,6 +165,14 @@
   tags:
     - search
 
+- name: Populate common facts
+  hosts: all
```
alxgomz, line 169: let's be specific I trhink we should just have `respository:acc:adw`. That will also avoid includuing external hosts.

playbooks/acs.yml
```diff
@@ -318,6 +326,16 @@
...
+- name: Populate adf app facts
+  hosts: acc:adw
+  gather_facts: false
+  tasks:
+    - name: Populate adf app config when identity available
+      when: identity_enabled
+      ansible.builtin.set_fact:
+        adf_app_oauth_configuration:
+          APP_CONFIG_AUTH_TYPE: oauth
+
```
alxgomz, line 338: So we rely on the fact the default config match the initial use case?
I think adding some more params to later on suport external identity is relatively cheap, isn't it?

alxgomz (changes requested):
at least change the `hosts` `identity_enabled` is set for

playbooks/acs.yml
gionn: yes, and I remember we need those defaults present because the packaged configuration contains $placeholders which must be replaced to have a working install (even when sso is disabled).

I think I could repeat here the properties which probably needs to be changed when using an external identity:
```
APP_CONFIG_IDENTITY_HOST: "{protocol}//{hostname}{:port}/auth/realms/alfresco"
APP_CONFIG_OAUTH2_HOST: "{protocol}//{hostname}{:port}/auth/realms/alfresco"
APP_CONFIG_OAUTH2_CLIENTID: alfresco
```

molecule/identity/verify.yml
gionn: OPSEXP-2327

alxgomz (approved):

PR #680, opened 2023-10-03
OPSEXP-2314 Enable SSO for Share

gionn:
Ref: OPSEXP-2314

### DIFF
--- a/roles/repository/tasks/share.yml
+++ b/roles/repository/tasks/share.yml
@@ -4,7 +4,7 @@
     dynamic_share_properties: {}
 
 - name: Configure identity service properties when available
-  when: repository_identity_url | length > 0
+  when: repository_identity_url
   vars:
     sso_share_properties:
       aims:
@@ -18,7 +18,7 @@
 
 - name: Prepare final structure for share properties
   ansible.builtin.set_fact:
-    merged_share_properties: "{{ default_share_properties | combine(dynamic_share_properties) }}"
+    merged_share_properties: "{{ default_share_properties | ansible.builtin.combine(dynamic_share_properties) }}"
 
 - name: Add the configuration templates
   ansible.builtin.template:
--- a/roles/repository/tasks/share.yml
+++ b/roles/repository/tasks/share.yml
@@ -1,0 +1,41 @@
+---
+- name: Initialize accumulator for dynamic share properties
+  ansible.builtin.set_fact:
+    dynamic_share_properties: {}
+
+- name: Configure identity service properties when available
+  when: repository_identity_url | length > 0
+  vars:
+    sso_share_properties:
+      aims:
+        enabled: "true"
+        realm: "{{ repository_identity_realm }}"
+        resource: "{{ repository_identity_client_id }}"
+        secret: "{{ repository_identity_client_secret }}"
+        authServerUrl: "{{ repository_identity_url }}"
+  ansible.builtin.set_fact:
+    dynamic_share_properties: "{{ dynamic_share_properties | ansible.builtin.combine(sso_share_properties) }}"
+
+- name: Prepare final structure for share properties
+  ansible.builtin.set_fact:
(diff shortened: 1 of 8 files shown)


roles/repository/tasks/share.yml
```diff
@@ -0,0 +1,41 @@
+---
+- name: Initialize accumulator for dynamic share properties
+  ansible.builtin.set_fact:
+    dynamic_share_properties: {}
+
+- name: Configure identity service properties when available
+  when: repository_identity_url | length > 0
```
alxgomz, line 7: I've seen we've spreaded that syntax in many different places lately (and not so lately).
Do we want make it a coding standard? I personnaly prefer the shorthand `repository_identity_url` which evaluate the same way as far as I can tell

roles/repository/tasks/share.yml
```diff
@@ -0,0 +1,41 @@
...
+        realm: "{{ repository_identity_realm }}"
+        resource: "{{ repository_identity_client_id }}"
+        secret: "{{ repository_identity_client_secret }}"
+        authServerUrl: "{{ repository_identity_url }}"
+  ansible.builtin.set_fact:
+    dynamic_share_properties: "{{ dynamic_share_properties | ansible.builtin.combine(sso_share_properties) }}"
+
+- name: Prepare final structure for share properties
+  ansible.builtin.set_fact:
+    merged_share_properties: "{{ default_share_properties | combine(dynamic_share_properties) }}"
```
alxgomz, line 21: fqcn filter name please.

roles/repository/tasks/share.yml
```diff
@@ -0,0 +1,41 @@
...
+  vars:
+    sso_share_properties:
+      aims:
+        enabled: "true"
+        realm: "{{ repository_identity_realm }}"
+        resource: "{{ repository_identity_client_id }}"
+        secret: "{{ repository_identity_client_secret }}"
+        authServerUrl: "{{ repository_identity_url }}"
+  ansible.builtin.set_fact:
+    dynamic_share_properties: "{{ dynamic_share_properties | ansible.builtin.combine(sso_share_properties) }}"
```
alxgomz, line 17: You're overiding  `dynamic_share_properties` with itself... From what I remeber this doesn't work great (but I could be mixing up with something else), but more importantly the next task creates a new fact `merged_share_properties` and do not use this "self override" thingy... Is there any reason for using different approaches in those 2 tasks?

alxgomz (changes requested):

roles/repository/tasks/share.yml
gionn: not sure, because in the end with `repository_identity_url: ''` still evaluates to false.

The other similar usages I see are related to secret where we could enforce a minimum length rather than plain `> 0`

roles/repository/tasks/share.yml
gionn: please recheck, usage is consistent.

more in general, this is an accumulator pattern which helps when there are more iteration which needs to extend the same structure. for now the only unexpected behavior I have seen it's only related to combine filter.

alxgomz: I should probably have placed this comment after line 21. Essentialy I'm wondering why we are not updating `dynamic_share_properties` again line 21 and instead creating another fact while in line 17 we are chosing to update the var. I guess it's not a big deal.

alxgomz (approved):

roles/repository/tasks/share.yml
gionn: ah, for proper precedence handling, I want that default_share_properties could be overridden by the dynamic_share_properties ones, not viceversa

gionn: ok maybe I understood only now, ideally I could have done:

```
    dynamic_share_properties: "{{ default_share_properties | ansible.builtin.combine(dynamic_share_properties) }}"
```

but I wanted to make explicit that after L21 any other change to dynamic_share_properties would not be reflected on the configuration template, hence I created a new variable

PR #681, opened 2023-10-03
OPSEXP-2290: fix sync service and transformation configs

alxgomz:
Ref: OPSEXP-2290
transformation config changed in  https://github.com/Alfresco/alfresco-enterprise-repo/pull/1359
Sync service role just needed to work from the distributed file rather than local templates


molecule/multimachine/host_vars/repository_neutron.yml
```diff
@@ -1,2 +1,2 @@
 # variables for molecule tests - inventory_hostname: repository_neutron
-cluster_keepoff: yes
+cluster_keepoff: false  # workaround while investigatinfclusering issues
```
gionn, line 2: raise a new issue and reference it there? or at least fix the typos 😀

gionn (changes requested):
I would prefer having those changes on master since there is no reason to gate them until next release (except for the version bumps of course)

alxgomz:
> I would prefer having those changes on master since there is no reason to gate them until next release (except for the version bumps of course)

This taskl being part of the 23.1 readiness epic the target branch is for now next/acs-23.1

gionn (approved):

PR #688, opened 2023-10-10
OPSEXP-2314 Make sure that share SSO config is properly setup

gionn:
* fixup redirect uris in the alfresco realm
* make sure that clients use the main alfresco fqdn to access keycloak

Ref: OPSEXP-2314

### DIFF
--- a/playbooks/acs.yml
+++ b/playbooks/acs.yml
@@ -28,9 +28,14 @@
 - name: Gather facts from all hosts
   hosts: all:!external
   gather_facts: false
+  roles:
+    - role: "../roles/common"
   tasks:
     - name: Setup
       ansible.builtin.setup:
+    - name: Set fqdn fact
+      ansible.builtin.set_fact:
+        alfresco_url: "{% if use_ssl | bool %}https{% else %}http{% endif %}://{% if fqdn_alfresco %}{{ fqdn_alfresco }}{% else %}localhost{% endif %}"
 
 - name: Database Role
   hosts: database
@@ -65,6 +70,20 @@
   tags:
     - elasticsearch
 
+- name: Nginx Role
+  hosts: nginx
+  gather_facts: false
+  roles:
+    - role: "../roles/nginx"
+  post_tasks:
+    - name: Make sure NGINX is running
+      become: true
+      ansible.builtin.service:
+        name: nginx
+        state: started
+  tags:
+    - nginx
+
 - name: Identity Role
   hosts: identity
   gather_facts: false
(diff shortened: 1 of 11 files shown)


playbooks/acs.yml
```diff
@@ -28,9 +28,14 @@
 - name: Gather facts from all hosts
   hosts: all:!external
   gather_facts: false
+  roles:
+    - role: "../roles/common"
   tasks:
     - name: Setup
       ansible.builtin.setup:
+    - name: Set fqdn fact
```
gionn, line 36: find a better name

playbooks/acs.yml
```diff
@@ -25,12 +25,15 @@
...
   hosts: all:!external
   gather_facts: false
+  roles:
+    - role: "../roles/common"
   tasks:
-    - name: Setup
-      ansible.builtin.setup:
+    - name: Set alfresco url from fqdn_alfresco if available or localhost
+      ansible.builtin.set_fact:
+        alfresco_url: "{% if use_ssl | bool %}https{% else %}http{% endif %}://{% if fqdn_alfresco %}{{ fqdn_alfresco }}{% else %}localhost{% endif %}"
```
alxgomz, line 36: Just because `fqdn_alfresco` is not set doesn't mean install is done on localhost. `fqdn_alfresco` is a param that was introduced for ssl configuration only and is documented as such.
Though, I understand what we want to do is not "conventional" and proper implementation would require more time. let's take some time to discuss it a bit further

gionn: I got inspired by https://github.com/Alfresco/alfresco-ansible-deployment/blob/f96645fe04025c7e2ee6891e7483b29929f3db4c/roles/repository/templates/alfresco-global.properties.j2#L57 - I could try to reintroduce `nginx_host` for consistency, I used localhost just as a shortcut knowing that would work locally and on cloud-infra where we have ssl setup in place.

To summarize a little, I need a fact for the main url where clients can access the application because it needs to be known for keycloak redirect uri and clients needs to use the same otherwise the oauth issuer will not match.

alxgomz: For this use case I would normally use `known_urls | first` but on cloud-infra that would mean an https url. Would it work with the current proxy setup?

alxgomz (dismissed):

playbooks/acs.yml
```diff
@@ -25,12 +25,15 @@
     secret_ca_passphrase: "{{ hostvars.localhost.ca_signing_key_passphrase }}"
   ansible.builtin.import_playbook: pki.yml
 
-- name: Gather facts from all hosts
-  hosts: all:!external
+- name: Setup common facts
+  hosts: repository:identity
```
alxgomz, line 29: That looks brutal. I don't think hosts without repo or identoity roles can complete without the common role being executed on them.
What theretional behind this change?

gionn: common role is getting applied everywhere since it's a dependency of any other role, but the problem here is that I want to reference facts before the execution of any role, hence the explicit dependency requirement

PR #689, opened 2023-10-11
OPSEXP-2335 Fixup app.config.json dynamic configuration for adf apps

gionn:
Ref: OPSEXP-2335


gionn:
it seems that this playbook has only been ever released with acc 8.0.0, so we should safely get rid of $placeholder replacement and go straight with merging json

roles/adf_app/molecule/default/verify.yml
alxgomz, line 1: can you also check the http redirect is working as intended?

PR #696, opened 2023-10-18
OPSEXP-2329 Fix keycloak endpoint and console access behind proxy

gionn:
Ref: OPSEXP-2329

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -1,3 +1,9 @@
+# REQUIRED: add at least the Share url
+# The first entry will also be used to assume the base URL of this alfresco install
+# For more information please have a look at the
+# [security_doc](https://github.com/Alfresco/alfresco-ansible-deployment/blob/master/docs/SECURITY.md#specify-trustworthy-applications)
+known_urls: []
+
 artifacts_repositories:
   enterprise:
     base_url: "https://artifacts.alfresco.com/nexus/content"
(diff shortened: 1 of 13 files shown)


group_vars/all.yml
```diff
@@ -1,3 +1,9 @@
+# REQUIRED: add at least the Share url
+# The first entry will also be used to assume the base URL of this alfresco install
+# For more information please have a look at the
+# [security_doc](https://github.com/Alfresco/alfresco-ansible-deployment/blob/master/docs/SECURITY.md#specify-trustworthy-applications)
+known_urls: []
```
alxgomz, line 5: I think we should set the localhost default here otehrwise I suspect the vagrant deploy will break given it has it set in the invetory file as a group var which has even lower precedence

alxgomz (changes requested):

group_vars/all.yml
gionn: need to decide how to proceed here (not really for vagrant but more about the breaking change of moving known_urls)

the only alternative I have in mind is that we don't use known_urls and provide an additional variable to be used only when necessary (identity enabled only for now?)

alxgomz: I think https://github.com/Alfresco/alfresco-ansible-deployment/commit/d982225e074256addc7543ac59d14c23f97f54a2 makes `known_urls` mandatory and documents the first entry is a special case.
I prefer we stick to that for now and cahnge the behaviour in 3.x

alxgomz (approved):

PR #697, opened 2023-10-18
OPSEXP-2238 Support for deploying with ansible on Rockylinux 9

slohe1:
OPSEXP-2238


docs/README.md
```diff
@@ -75,6 +75,7 @@ Legend:
 | RHEL 8.2            | -                        | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
 | RHEL 7.7            | -                        | -                  | -                  | :white_check_mark: | :white_check_mark: |
 | RHEL 7.6            | -                        | -                  | -                  | :white_check_mark: | :white_check_mark: |
+| Rockylinux 9.2      | :white_check_mark:       | -                  | -                  | -                  | -                  |
```
gionn, line 78: add also rockylinux 8.7 for the previous one

group_vars/all.yml
```diff
@@ -86,6 +86,7 @@ supported_os:
       - 7.6
   Rocky:
     versions:
+      - 9.2
       - 8.7
       - 8.6
```
gionn, line 91: still not sure what we should do here but ok

roles/java/tasks/keystores.yml
```diff
@@ -28,6 +28,11 @@
         mode: "0400"
         dest: "{{ download_location }}"
       register: certcp
+    - name: Assign variable based on distribution
```
gionn, line 31:
```suggestion
    - name: Add legacy parameter for compatibility with newer openssl versions
```

roles/java/tasks/keystores.yml
```diff
@@ -28,6 +28,11 @@
...
+      ansible.builtin.set_fact:
+        legacy: "-legacy"
```
gionn, line 33: call the variable something like `additional_options` so it's more explicit what we are doing

roles/java/tasks/keystores.yml
```diff
@@ -28,6 +28,11 @@
...
+      when:
+        - (ansible_distribution == 'Rocky' or ansible_distribution == 'Redhat') and ansible_distribution_version is version('9', '>=')
```
alxgomz, line 35: Isn't `ansible_os_family` set to `RedHat` with Rocky?
Also AFAIR `RedHat` needs a capital "H". Please double check.

alxgomz (changes requested):
You're not installing curl for ReHat based distros in the repository role

alxgomz (changes requested):

roles/repository/tasks/main.yml
```diff
@@ -41,6 +41,25 @@
...
+
+- name: Install 'curl' on RedHat 7
+  become: true
+  ansible.builtin.yum:
+    name: curl
+    state: present
+  when:
+    - ansible_os_family == 'RedHat'
+    - ansible_distribution_major_version == '7'
+
```
gionn, line 62: seems too complex - the point is that on recent RHEL a curl-minimal [began to ship into base image](https://fedoraproject.org/wiki/Changes/CurlMinimal_as_Default#Summary) because it's a requirement for dnf and it's fine that we just use curl-minimal until there are clear issues with it.

having a task which use the generic package module to install `curl` everywhere except on redhat >= 9 should be fine

gionn (changes requested):

roles/repository/tasks/main.yml
```diff
@@ -41,6 +41,15 @@
...
       when: temp_array|length > 0
 
+- name: Install curl except Redhat 9
+  become: true
+  ansible.builtin.package:
+    name: curl
+    state: present
+  when:
+    - ansible_os_family == 'RedHat'
+    - ansible_distribution_major_version is not version('9', '>=')
```
gionn, line 51:
```suggestion
    - ansible_distribution_major_version is version('9', '<')
```

to make De Morgan happy

alxgomz (approved):
I still think we should always try to install curl and only go green if the error reports conflict with curl-minimal.
Though, I guess we can go with that solution for now and revisit in case someone comes to us with a redhat9 based system that doesn't have curl-minimal ootb.

PR #698, opened 2023-10-18
OPSEXP-2354 Fix nginx redirect when behind ELB via argument

gionn:
Ref: OPSEXP-2354

### DIFF
--- a/roles/nginx/meta/argument_specs.yml
+++ b/roles/nginx/meta/argument_specs.yml
@@ -11,10 +11,10 @@ argument_specs:
         type: bool
         default: true
       nginx_set_proxy_headers:
-        description: Useful to disable when nginx node is behind another reverse proxy (e.g. AWS ELB)
+        description: Useful when nginx node is behind another reverse proxy (e.g. should be disabled when behind an AWS ELB)
         type: bool
         default: true
       nginx_absolute_redirect:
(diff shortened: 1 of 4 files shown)


roles/nginx/meta/argument_specs.yml
```diff
@@ -10,3 +10,11 @@ argument_specs:
         description: If the nginx reverse proxy vhosts should be enabled
         type: bool
         default: true
+      nginx_set_proxy_headers:
+        description: Useful to disable when nginx node is behind another reverse proxy (e.g. AWS ELB)
```
alxgomz, line 14:
```suggestion
        description: Useful when nginx node is behind another reverse proxy (e.g. should be disabled when behind an AWS ELB)
```

roles/nginx/meta/argument_specs.yml
```diff
@@ -10,3 +10,11 @@ argument_specs:
...
+        type: bool
+        default: true
+      nginx_absolute_redirect:
+        description: Useful to disable when nginx node is behind another reverse proxy (e.g. AWS ELB)
```
alxgomz, line 18:
```suggestion
        description: Useful when nginx node is behind another reverse proxy (e.g. should be disabled when behind an AWS ELB)
```

alxgomz (approved):

slohe1 (approved):

PR #703, opened 2023-10-25
OPSEXP-2156 Bump molecule to v6.0.3

gionn:
* Bump pre-commit hooks
* Pin pre-commit python version to the main playbook python version

Ref: OPSEXP-2156


alxgomz (approved):
I hope it wasn't to boring :)

PR #710, opened 2023-11-15
Allow access to Control Center for Community

morgan-patou:
Alfresco Control Center is installed even for Community but it is only accessible from localhost:8881 because of Security Filters and other things. So it is necessary to add the proxy from the main 80/443 ports into the server hosting this ADF-based application so that it can be accessed.

For example, if you try to install a simple Community version, with the current template, you will have a /auth/ URL that will fail if there is no Identity (KeyCloak) component and you won't have access to ACC.

I believe it would be better to add conditions for Identity URL (a condition like `identity` group not empty?) and to remove the condition for ACC (this PR or even better, have a condition like `acc` group not empty?).

### DIFF
--- a/roles/nginx/templates/alfresco_proxy.include.j2
+++ b/roles/nginx/templates/alfresco_proxy.include.j2
@@ -49,6 +49,11 @@
         include {{ nginx_vhost_path }}/alfresco_proxy_headers.include;
     }
 
+    location /control-center/ {
+        proxy_pass http://{{ acc_host }}:8881/;
+        include {{ nginx_vhost_path }}/alfresco_proxy_headers.include;
+    }
+
 {% if acs.edition == "Enterprise" %}
     location /syncservice/ {
         proxy_pass http://{{ sync_host }}:{{ ports_cfg.sync.http }}/alfresco/;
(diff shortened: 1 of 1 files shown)


alxgomz (approved):

morgan-patou:
I added another commit that, for security reasons (and also to avoid errors on URLs), will only add the nginx location into the config file if the components are really installed (i.e. the group exists and isn't empty). I did that for the main optional components (identity, acc, adw, syncservice) and I left the rest that will be installed always (repo, share).

I tested the generation of the config file in all cases (enterprise, community, with all the groups, with some groups, with no groups) and it all looked fine to me.

roles/nginx/templates/alfresco_proxy.include.j2
```diff
@@ -44,24 +44,32 @@
         include {{ nginx_vhost_path }}/alfresco_proxy_headers.include;
     }
 
+{% if groups['identity'] | default([]) | length != 0 %}
```
gionn, line 47: > will only add the nginx location into the config file if the components are really installed

thanks, something which is good to have!

Would you mind changing the length checks to `length > 0` which should more consistent?

morgan-patou: No problem, it's done. Sometimes the `groups['repository']` is used and sometimes it's `groups.repository`. Not sure if you would like to align all of them as well at some point.

gionn: I don't have strong opinions on this but better be consistent

gionn (approved):

gionn:
Thanks for the contribution!

PR #635, opened 2023-06-30
Hash inventory_hostname when used in path to avoid special characters

morgan-patou:
As discussed in issue #623, it would be useful to hash the `inventory_hostname` when it is used for path, to make sure that no special characters come into play and cause issues. In this PR, I changed the paths as well as the SSL Certificate alias because aliases have the same restrictions as hostnames.

Note 1: I didn't touch the `molecule` folders since this is your internal testing - let me know if you would like me to change them as well

Note 2: I didn't change `roles/common/tasks/check_upgrades.yml` as well, since the path `/tmp/{{ inventory_hostname }}/{{ ansible_installation_status_file }}` doesn't appear to have any other references anywhere at of this time. Therefore I would assume this was used in the past releases to store the status file but not anymore (`"{{ binaries_folder }}/.ansible_alfresco_components.status"`)

List of all inventory_hostname:
```bash
mop@DBI-LT-MOP2:alfresco-ansible-deployment$ grep -ri --color inventory_hostname *
docs/deployment-guide.md:* an `ansible_host` variable; if the host needs to be reached through an address that's different from the `inventory_hostname` (e.g. machine is only reachable through a bastion host or some sort of NAT).
molecule/multimachine/host_vars/repository_electron.yml:# variables for molecule tests - inventory_hostname: repository_electron
molecule/multimachine/host_vars/repository_proton.yml:# variables for molecule tests - inventory_hostname: repository_proton
molecule/multimachine/host_vars/repository_neutron.yml:# variables for molecule tests - inventory_hostname: repository_neutron
molecule/pki/verify.yml:        src: ../../configuration_files/pki/certificates/{{ inventory_hostname }}.p12
molecule/pki/verify.yml:        path: /tmp/{{ inventory_hostname }}.crt
molecule/pki/verify.yml:        src: /tmp/{{ inventory_hostname }}.crt
molecule/pki/verify.yml:          client_crt: /tmp/{{ inventory_hostname }}.crt
playbooks/pki.yml:        path: "{{ hostvars.localhost.pki_dir }}/certificates/{{ inventory_hostname }}.p12"
playbooks/pki.yml:            path: /etc/pki/{{ inventory_hostname }}_{{ cert_key_type | default('') }}.key
playbooks/pki.yml:              - DNS:{{ inventory_hostname }}
playbooks/pki.yml:            common_name: "{{ ansible_fqdn | default(inventory_hostname) }}"
[...]

Issue #623, opened 2023-06-20
Limit the use of inventory_hostname


morgan-patou:
## Summary
`inventory_hostname` might not be the real hostname of the target host and might contain characters that aren't supported by hostnames or that can cause issues with path.

## Details
As you know, `inventory_hostname` comes from the inventory and therefore, I believe it's a kind of free-text name, which can contain characters such as `/`. Recently, a lot of addition was done on the Java/PKI side and this variable is being used a lot more. In cases where such characters are being used in the inventory_hostname, then some of the commands added recently would fail.

For example:
```bash
        - name: Create private key for new certificate
          no_log: true
          become: true
          community.crypto.openssl_privatekey:
            path: /etc/pki/{{ inventory_hostname }}_{{ cert_key_type | default('') }}.key
            mode: 0600
            size: "{{ cert_key_size | default(omit) }}"
            type: "{{ cert_key_type | default(omit) }}"
            return_content: true
          register: srvkey
```

Because `inventory_hostname` is being used inside the value of a path, if this variable starts with `infra/myhost`, then this task would fail because the folder `/etc/pki` exists but not `/etc/pki/infra/`.

Hostname are following the regex `[a-zA-Z0-9.-]*` and therefore inventory_hostname might not be the real value.

For all the Search Replication role, you are also using `inventory_hostname` to define if the current host is a master or a slave with things such as `{{ 'slave' if inventory_hostname == search_master else 'master' }}` or `{% if search_topology == 'replication' and inventory_hostname == search_master %}`. I find these conditions to be quite strange, since that would mean that you are forcing the inventory_hostname to be exactly "search_master" no? Maybe I'm missing some things for the replication part.

## Questions
Is it possible to change such occurrences of `inventory_hostname` so that it uses the real hostname which I believe would be more secure in terms of allowed characters? It could mean replacing most occurrences with `ansible_hostname` or `ansible_facts['hostname']` for example (these two requires `gather_facts: true`). If the real hostname changes (not the inventory one), the SSL Certificate should be regenerated I assume, so using the inventory one might be a problem as well no?

[...]

PR #711, opened 2023-11-16
Create pki subfolders for inventory hosts with slash

morgan-patou:
As discussed in issue #623 and then PR #635, I'm creating another PR (the 635 was too far behind, too many changes could come into conflict) that is for subfolder creation in the PKI playbook.

This updated version will create the needed subfolders on both the remote server and the Ansible Control Node as needed (and also fixing a few typos). In the end, it's only updating the line 161 to include any possible directory coming from the `inventory_hostname` variable. The `| dirname` takes care of everything so even if there is no `/` in the name, then the /etc/pki folder gets created (if it doesn't exist yet). And if it does contain one, then it will create both the /etc/pki and any subfolder needed.

I also added a task to create the folder on the Control Node around the end of the file, since this is also needed and I don't think I can put it at the beginning of the file because I don't know all the `inventory_hostname` for all hosts at this point in time.

Note/Question: what is the use of creating the `- name: /etc/pki/private` folder on the target server (line 162/163)? From what I can see, this folder is never used on the target server, is it?

Thanks!
Morgan

### DIFF
--- a/playbooks/pki.yml
+++ b/playbooks/pki.yml
@@ -209,6 +209,7 @@
           ansible.builtin.file:
             path: "{{ hostvars.localhost.pki_dir }}/certificates/{{ inventory_hostname | dirname }}"
             state: directory
+            mode: "0750"
           delegate_to: localhost
 
         - name: Sign certificate with our CA
--- a/playbooks/pki.yml
+++ b/playbooks/pki.yml
@@ -171,13 +171,15 @@
             state: present
 
         - name: Create private key for new certificate
+          vars:
+            key_type: "{{ cert_key_type | default('RSA') }}"
           no_log: true
           become: true
           community.crypto.openssl_privatekey:
-            path: /etc/pki/private/{{ inventory_hostname | basename }}_{{ cert_key_type | default('') }}.key
+            path: /etc/pki/private/{{ inventory_hostname | basename }}_{{ key_type }}.key
             mode: 0600
             size: "{{ cert_key_size | default(omit) }}"
-            type: "{{ cert_key_type | default(omit) }}"
+            type: "{{ key_type }}"
             return_content: true
           register: srvkey
 
--- a/playbooks/pki.yml
+++ b/playbooks/pki.yml
@@ -158,9 +158,9 @@
             state: directory
             mode: "{{ item.mode | default(omit) }}"
           loop:
-            - name: /etc/pki/{{ inventory_hostname | dirname }}
+            - name: /etc/pki
             - name: /etc/pki/private
-              mode: "0600"
+              mode: "0700"
 
         - name: Install PKI required packages
           become: true
@@ -174,7 +174,7 @@
           no_log: true
           become: true
           community.crypto.openssl_privatekey:
-            path: /etc/pki/{{ inventory_hostname }}_{{ cert_key_type | default('') }}.key
+            path: /etc/pki/private/{{ inventory_hostname | basename }}_{{ cert_key_type | default('') }}.key
             mode: 0600
             size: "{{ cert_key_size | default(omit) }}"
             type: "{{ cert_key_type | default(omit) }}"
--- a/playbooks/pki.yml
+++ b/playbooks/pki.yml
@@ -13,8 +13,8 @@
           - p12_passphrase
           - secret_ca_passphrase
         fail_msg: |
-          this playbook do not generate passwords. You must generate then and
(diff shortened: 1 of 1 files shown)


playbooks/pki.yml
```diff
@@ -157,7 +158,7 @@
             state: directory
             mode: "{{ item.mode | default(omit) }}"
           loop:
-            - name: /etc/pki
+            - name: /etc/pki/{{ inventory_hostname | dirname }}
             - name: /etc/pki/private
               mode: "0600"
```
alxgomz, line 163: this mode looks inappropriate. Do you mind fixing it while you're doing that?

playbooks/pki.yml
```diff
@@ -157,7 +158,7 @@
             state: directory
             mode: "{{ item.mode | default(omit) }}"
           loop:
-            - name: /etc/pki
+            - name: /etc/pki/{{ inventory_hostname | dirname }}
```
alxgomz, line 161: Here I would not create subdirs and use `inventory_hostname | basename` when generating the key instead. This path is not reused anywhere else IIRC and thats rather unusual to have subdirs on the target host pki directory (even more becaue those subdir probably wil not make any sense at the host level)

alxgomz (changes requested):

alxgomz:
Regarding your question: That's a valid point. The key should actually be generated in the `/etc/pki/private` folder not in its parent (nothing critical I guess but that would be more inline with "common practices").

playbooks/pki.yml
morgan-patou: I changed it to 700. Let me know if you prefer 750 instead. Depending on distribution, it's either 700 or 750 for /etc/pki or /etc/ssl.

playbooks/pki.yml
morgan-patou: I reverted that part and use the basename on key creation.

morgan-patou:
> Regarding your question: That's a valid point. The key should actually be generated in the `/etc/pki/private` folder not in its parent (nothing critical I guess but that would be more inline with "common practices"). You're very welcome to fix it :)

Yes exactly. I put the key into the private folder. I will verify everything is working for me with the last version but it should be good.

playbooks/pki.yml
```diff
@@ -173,7 +174,7 @@
           no_log: true
           become: true
           community.crypto.openssl_privatekey:
-            path: /etc/pki/{{ inventory_hostname }}_{{ cert_key_type | default('') }}.key
+            path: /etc/pki/private/{{ inventory_hostname | basename }}_{{ cert_key_type | default('') }}.key
```
morgan-patou, line 177: Could you please let me know if you would be fine with this alternative line:
```yml
        path: "/etc/pki/private/{{ inventory_hostname | basename }}{% if cert_key_type %}_{{ cert_key_type }}{% endif %}.key"
```

Currently, if `cert_key_type` isn't defined, then the key name will end with `_.key`, which isn't very pretty. What do you think about removing this underscore in case no key type is provided?

alxgomz: I'd rather add a task var to make sure the `type` module param is always set.

morgan-patou: So your approach would be to use the following instead:
```yml
          community.crypto.openssl_privatekey:
            path: /etc/pki/private/{{ inventory_hostname | basename }}_{{ cert_key_type | default('RSA') }}.key
            mode: 0600
            size: "{{ cert_key_size | default(omit) }}"
            type: "{{ cert_key_type | default('RSA') }}"
            return_content: true
```

Isn't that more dangerous that anything else to force a `type` to be present and not omitted by default? If we force a default to "RSA"(or something else), then it might break at some point in the future, if this type is removed.

alxgomz: Yes that's what I had in mind except I would have used a task level var to a

```yaml
- name: ....
  vars:
    key_type: "{{ cert_key_type | default('RSA') }}"
  community.crypto.openssl_privatekey:
    path: /etc/pki/private/{{ inventory_hostname | basename }}_{{ key_type }}.key
    mode: 0600
    size: "{{ cert_key_size | default(omit) }}"
    type: "{{ key_type }}"
    return_content: true
```

morgan-patou: It was added

alxgomz (approved):

alxgomz:
Thanks for this new contrib! 🙏

PR #713, opened 2023-11-17
OPSEXP-2399: move pki molecule scenario to comunity workflow

alxgomz:
Ref: OPSEXP-2399

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,72 @@
+name: "docker integration"
+
+on:
+  workflow_call:
+    inputs:
+      scenario:
+        type: string
+        description: molecule scenario
+        required: false
+        default: "false"
+      os_distribution:
+        type: string
+        description: OS image to use for the container
+        required: false
+        default: rockylinux:9
+      dtas_version:
+        type: string
+        description: git ref to use when cloning DTAS repo
+        required: false
+        default: master
+    secrets:
+      nexus_username:
+        required: false
+      nexus_password:
+        required: false
+      dtas_token:
+        required: false
+
+env:
+  DTAS_VERSION: ${{ inputs.dtas_version }}
+  BUILD_NUMBER: ${{ github.run_id }}
+  PY_COLORS: 1
+  PYTHONUNBUFFERED: 1
+
+jobs:
+  docker_integration:
+    name: Test ${{ inputs.scenario }} scenario on ${{ inputs.os_distribution }}
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/docker.yml
```diff
@@ -0,0 +1,72 @@
...
+
+      - name: Install python packages
+        uses: Alfresco/alfresco-build-tools/.github/actions/pipenv@v3.4.0
+        with:
+          python-version: ${{ steps.setup-python.outputs.python-version }}
+
+      - name: Install and cache ansible galaxy dependencies
+        uses: ./.github/actions/galaxy
+        with:
+          cache-name: enterprise
```
gionn, line 61: mmm switch to another cache name to avoid poisoning between community and enterprise jobs

.github/workflows/enteprise.yml
```diff
@@ -78,55 +84,29 @@ jobs:
...
-          MOLECULE_ROLE_IMAGE: ${{ matrix.molecule_distro.image }}
-          NEXUS_USERNAME: ${{ secrets.nexus_username }}
-          NEXUS_PASSWORD: ${{ secrets.nexus_password }}
-          CLONE_GITHUB_TOKEN: ${{ secrets.BOT_GITHUB_TOKEN }}
-        run: pipenv run molecule test -s ${{ matrix.scenario.name }}
+    uses: ./.github/workflows/docker.yml
+    with:
+      scenario: ${{ matrix.scenario.name }}
+      os_distribution: ${{ matrix.molecule_distro.image }}
+      dtas_version: ${{ needs.docker.outputs.dtas_version }}
```
gionn, line 105: it's defined as env in the global level, you should be able to reference it here without the need of setting an output

gionn (changes requested):

.github/workflows/docker.yml
alxgomz: Why do you think it would poison the cache? the galaxy assets are identical for both ent. & com.

.github/workflows/enteprise.yml
alxgomz: the env context is not available in `jobs.<job_id>.with`

.github/workflows/docker.yml
gionn: I am talking about malicious stuff like the one described in https://scribesecurity.com/blog/github-cache-poisoning/

in general is better to keep separate caches for workflows which runs from forked repo and ours which has access to secrets - at the cost of increased cache usage which should not really have a big impact for us here

gionn (approved):

PR #719, opened 2023-11-21
Remove not really needed constraints in workflows

gionn:
* docker_integration waiting for docker in community is just making overall execution time longer
* docker_integration in enterprise can run on dependabot (nexus and github secrets are available)

### DIFF
--- a/.github/workflows/community.yml
+++ b/.github/workflows/community.yml
@@ -85,8 +85,6 @@ jobs:
   docker_integration:
     name: Test ${{ matrix.scenario.name }} scenario on ${{ matrix.molecule_distro.image }}
     if: github.actor != 'dependabot[bot]'
-    needs:
-      - docker
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


gionn:
> I would have done the opposite (allow dependabot PR in comunity but not in enterprise).

the only reason for skipping dependabot is when the workflow requires secrets, but secrets needed there are already available, so I don't see why skipping it

> Also why are the roles tests needed in ent. and not in comunity?

Not sure what you mean, the other change I did is because strictly speaking there is no need to execute docker integration tests after basic role tests.

alxgomz:
> > I would have done the opposite (allow dependabot PR in comunity but not in enterprise).
> 
> the only reason for skipping dependabot is when the workflow requires secrets, but secrets needed there are already available, so I don't see why skipping it
> 
Are you sure? It seems to me aws secrets are not available to dependabot.

> > Also why are the roles tests needed in ent. and not in comunity?
> 
> Not sure what you mean, the other change I did is because strictly speaking there is no need to execute docker integration tests after basic role tests.

I mean that the dependency `needs["docker"]` is still present in enterprise workflow.

alxgomz (approved):

Issue #716, opened 2023-11-21
Postgres Deployment post-task fails when external database is used


jalvarezferr:
The post task is missing a `when` clause, so it is run even when Postgres is not installed because `repo_db_url` or `sync_db_url` where provided.

https://github.com/Alfresco/alfresco-ansible-deployment/blob/ecc58887a6072b5613ccd16048a2de1e4c60297a/playbooks/acs.yml#L44

alxgomz:
internally tracked in OPSEXP-2407

alxgomz:
The error actually triggers as you probably still have a `database` host set in your inventory file while also providing external urls.

gionn:
@jalvarezferr can you confirm that removing every host from the `database` group in your inventory solve the issue?

alxgomz:
Delayed fix to OPSEXP-2408

morgan-patou:
I was planning to test the external DB today and I just found your issue so please allow me to participate here.

My understanding of the external DB is the following one: when defining yourself `repo_db_url`, the playbook is supposed to not use the "internal" PG but instead use the following properties to setup Alfresco:
```bash
repo_db_name: "alfresco"
repo_db_username: "alfresco"
repo_db_driver: "org.postgresql.Driver"
repo_db_url: "jdbc:postgresql://<host>:<port>/{{ repo_db_name }}"
```

When using this `repo_db_url`, you are indeed supposed to not have any host as part of the group `database`, otherwise there could be some side effects. The role for PG installation will still be executed if `sync_db_url` is empty (default), even if there is no syncservice group member, which isn't optimal. I quickly checked above #720 link and it seems that this was taken into account but isn't yet in the master branch. So that means for now, you should either not have anything in the `database` group OR define both `repo_db_url` AND `sync_db_url` to something not empty, otherwise the postgres role will be executed.


I tested on my side and for me, it works properly when the `database` role is empty with the following properties defined:
```bash
repo_db_host: "DB_HOST_VIP"
repo_db_port: "5432"
repo_db_name: "alfresco"
repo_db_username: "alfresco"
repo_db_driver: "org.postgresql.Driver"
repo_db_url: "jdbc:postgresql://{{ repo_db_host }}:{{ repo_db_port }}/{{ repo_db_name }}"

sync_db_host: "{{ repo_db_host }}"
sync_db_port: "{{ repo_db_port }}"
sync_db_name: "alfresco-sync"
sync_db_username: "alfresco-sync"
[...]

PR #720, opened 2023-11-21
OPSEXP-2407: add when condition in post_tasks for consistency

alxgomz:
Ref: OPSEXP-2395

aims at "fixing" #716

### DIFF
--- a/playbooks/acs.yml
+++ b/playbooks/acs.yml
@@ -38,11 +38,14 @@
 - name: Database Role
   hosts: database
   gather_facts: false
+  vars:
+    test_env: "{{ repo_db_url == '' or (groups.syncservice and sync_db_url == '') }}"
   roles:
     - role: "../roles/postgres"
-      when: repo_db_url == "" or sync_db_url == ""
+      when: test_env
   post_tasks:
     - name: Make sure PostgreSQL is running
+      when: test_env
       ansible.builtin.service:
         name: "{{ postgresql_service }}"
         state: started
@@ -268,6 +271,7 @@
       when: acs.edition == "Enterprise"
   post_tasks:
     - name: Update installation status file with Trouter
+      when: acs.edition == "Enterprise"
       become: true
       vars:
         trouter_components:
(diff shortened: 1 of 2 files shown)


playbooks/acs.yml
```diff
@@ -38,11 +38,14 @@
...
   roles:
     - role: "../roles/postgres"
-      when: repo_db_url == "" or sync_db_url == ""
+      when: test_env
   post_tasks:
     - name: Make sure PostgreSQL is running
+      when: test_env
       ansible.builtin.service:
         name: "{{ postgresql_service }}"
         state: started
```
gionn, line 51: we could also say that if you have external database then you should not have any host in the database group

a better fix would be to adopt the same external_something approach we are using for other components so become more natural having to deal with inventory to define the infrastructure, but for now looks fine to me telling to drop the host from the database group

playbooks/acs.yml
```diff
@@ -268,6 +271,7 @@
       when: acs.edition == "Enterprise"
   post_tasks:
     - name: Update installation status file with Trouter
+      when: acs.edition == "Enterprise"
```
gionn, line 274: were they leaving an empty file with community or something?

playbooks/acs.yml
alxgomz: Couldn't agree more on aligning with other external services approach, though that's an heavier refactoring we can only plan for a major bump.
I have created OPSEXP-2408 for that (I was sure we already had a ticket for that but couldn't find it).
Are you suggesting we should not fix it and simply advertise one should make sure no database host remain in the inventoryif the repo_db_url are specified?

playbooks/acs.yml
alxgomz: No actually it was setting the trouter version set in `group_vars/all.yml` as it is a root element and the community extra vars file is not nulling it.

playbooks/acs.yml
gionn: yes we should not fix it, we are adding additional conditional logic here which is already implemented by `hosts` and the desired behaviour would be obtained by not defining the host in the database group, so for me it's more a misconfiguration than an actual bug.

gionn: ⬆️

gionn (changes requested):

alxgomz:
superseeded by #721

PR #732, opened 2023-12-13
OPSEXP-2424: update OS matrix

alxgomz:
Ref: OPSEXP-2424

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -60,37 +60,13 @@ set of instructions.
 
 ## Versioning
 
-While ACS supports a wide range of OS, the playbook is only known to work and is
-supported for a subset of them. The table below gives detailed information on
-the status of supported OS.
+To understand which Operating System the Alfresco Ansible playbook can work
+against (on the target nodes), please refer to the [Supported
+platform doc](https://docs.alfresco.com/content-services/latest/support/) (pick
+the right Alfresco version).
 
-Legend:
-
-* :white_check_mark: platform supported and tested on this playbook
-* :x: platform not supported by this playbook (despite officially supported by Alfresco)
-* `-` platform not officially supported by Alfresco
-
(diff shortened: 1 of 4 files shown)


group_vars/all.yml
```diff
@@ -94,6 +97,9 @@ supported_os:
   Rocky:
     versions:
       - 9.2
+      - 9.1
+      - 9.0
+      - 8.8
```
gionn, line 102: I remember we specifically agreed to not add them since they were untested - and I was for adding them because 99% are working fine

docs/README.md
```diff
@@ -73,20 +73,21 @@ Legend:
...
+| RHEL 9              | :white_check_mark:        | -                  | -                  | -                  | -                  | -                  |
+| RHEL 8.8            | :white_check_mark:        | -                  | -                  | -                  | -                  | -                  |
+| RHEL 8.7            | -                         | :white_check_mark: | -                  | -                  | -                  | -                  |
+| RHEL 8.6            | -                         | :white_check_mark: | :white_check_mark: | -                  | -                  | -                  |
+| RHEL 8.5            | -                         | -                  | :white_check_mark: | :white_check_mark: | -                  | -                  |
+| RHEL 8.4            | -                         | -                  | -                  | :white_check_mark: | :white_check_mark: | :white_check_mark: |
+| RHEL 8.2            | -                         | -                  | -                  | :white_check_mark: | :white_check_mark: | :white_check_mark: |
```
gionn, line 83: I am a bit worried that in the end here we are mirroring the docs of the standard supported platforms, except for the OS we don't explicitly support.

Can't we just say the playbook don't support Amazon Linux and Suse and please refers to the supported platform page to find the supported os version?

alxgomz: Sounds better indeed.
@anxumalo We will remove this table to avoid discrepencies as we are pretty bad at maintaining it. We'll just mention the exceptions we have compare to what's in docs. Might need an update on Ansible docs as well

gionn (approved):

PR #757, opened 2024-01-17
OPSEXP-2417 Disable APS integration in the ADW

slohe1:


alxgomz (dismissed):

roles/adf_app/defaults/main.yml
```diff
@@ -25,3 +25,6 @@ adf_app_default_configuration:
     redirectSilentIframeUri: "{protocol}//{hostname}{:port}/assets/silent-refresh.html"
     redirectUri: "/"
     redirectUriLogout: "/"
+  plugins:
+    processService: false
+    microsoftOnline: false
```
gionn, line 30: this is adw specific, should be set as `adf_app_configuration` at the playbook level, otherwise it get set also for `acc` which doesn't implement any of this.

gionn (changes requested):

playbooks/acs.yml
```diff
@@ -406,7 +411,7 @@
         {{ lookup('url', downloads.adw_zip_sha1_checksum_url, username=nexus_user, password=nexus_password) }}
         {%- endif %}
       adf_app_version: "{{ adw.version }}"
-      adf_app_configuration: "{{ adf_app_oauth_configuration | default({}) }}"
+      adf_app_configuration: "{{ adf_app_oauth_configuration | default({}) | ansible.builtin.combine(adf_plugins_configuration) }}"
```
gionn, line 414:
```suggestion
      adf_app_configuration: "{{ adf_plugins_configuration | ansible.builtin.combine(adf_app_oauth_configuration | default({}), recursive=true) }}"
```

@alxgomz please check

gionn (changes requested):

playbooks/acs.yml
alxgomz: I can't think of cases where the two vars would overlap. And both are task level static vars, so both would work.

alxgomz (approved):

gionn (approved):

PR #783, opened 2024-02-16
OPSEXP-2450 Search Enterprise as default search engine

gionn:
Ref: OPSEXP-2450


pmacius (approved):

alxgomz (changes requested):
small spelling mistake

alxgomz (approved):

PR #793, opened 2024-02-27
OPSEXP-2510 Bump versions for Eridanus release

gionn:
Ref: OPSEXP-2510


pmacius (approved):

community-extra-vars.yml
```diff
@@ -1,28 +1,28 @@
...
-  version: 2.0.8.2
+  version: 2.0.9.1
 amps:
   aos_module:
     repository: "{{ nexus_repository.releases }}/aos-module/alfresco-aos-module"
-    version: 2.0.0
+    version: 3.0.0
   googledrive_repo:
     repository: "{{ nexus_repository.releases }}/integrations/alfresco-googledrive-repo-community"
     version: 4.0.0
```
alxgomz, line 28: should be 4.1
Also looks like the share AMP seem to point to the repo one

Also bump in `group_vars/all.yml`

alxgomz (changes requested):

community-extra-vars.yml
gionn: >  Also looks like the share AMP seem to point to the repo one

aren't they following the same release cadence?

alxgomz: They do yes but the artifacts name for share is also `alfresco-googledrive-repo-community`

alxgomz (approved):

PR #803, opened 2024-03-01
OPSEXP-2537 Set credentials only if needed for current inventory

pmacius:
!Ref OPSEXP-2537

### DIFF
--- a/playbooks/secrets-init.yml
+++ b/playbooks/secrets-init.yml
@@ -52,21 +52,37 @@
             gs_mode: "plugin"
           when: vault_init | default('') == 'plugin'
 
+        - name: Generate loop of mandatory passwords
+          ansible.builtin.set_fact:
+            password_loop:
+              - ca_signing_key_passphrase
+              - certs_p12_passphrase
+              - repo_db_password
+              - sync_db_password
+              - reposearch_shared_secret
+
+        - name: Append password_loop when activemq_password is needed
+          ansible.builtin.set_fact:
+            password_loop: "{{ password_loop + ['activemq_password'] }}"
+          when: ((groups.activemq | default([])) + (groups.external_activemq | default([]))) | length > 0
+
+        - name: Append password_loop when elasticsearch_password is needed
+          ansible.builtin.set_fact:
+            password_loop: "{{ password_loop + ['elasticsearch_password'] }}"
+          when: ((groups.elasticsearch | default([])) + (groups.external_elasticsearch | default([]))) | length > 0
+
+        - name: Append password_loop when identity_admin_password is needed
+          ansible.builtin.set_fact:
+            password_loop: "{{ password_loop + ['identity_admin_password'] }}"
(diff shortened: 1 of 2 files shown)


playbooks/secrets-init.yml
```diff
@@ -67,6 +63,40 @@
...
             - identity_admin_password
+          when: groups['all'] | length == 0
+
+        - name: Generate loop of mandatory passwords
+          ansible.builtin.set_fact:
+            password_loop:
+              - ca_signing_key_passphrase
+              - certs_p12_passphrase
+              - repo_db_password
+              - sync_db_password
```
gionn, line 74: can be handled via syncservice group check

playbooks/secrets-init.yml
```diff
@@ -67,6 +63,40 @@
...
+              - reposearch_shared_secret
```
gionn, line 75: can be handled via search group check

playbooks/secrets-init.yml
```diff
@@ -52,13 +52,9 @@
...
-          changed_when: true
-          when: not lookup('vars', item, default=False)
-          tags:
-            - molecule-idempotence-notest
-          loop:
+        - name: Generate loop of mandatory passwords when empty inventory is provided
+          ansible.builtin.set_fact:
+            password_loop:
             - ca_signing_key_passphrase
             - certs_p12_passphrase
```
alxgomz, line 59: I'm not sure in which case would we get an empty inventory? AFAIR our playbook is (unfortunately) very much tied to an inventory with the required structure of groups

pmacius: Secrets-init is used without inventory in pipeline test scenarios:
https://github.com/Alfresco/alfresco-ansible-deployment/blob/master/tests/molecule_it/script.sh#L8

alxgomz (dismissed):

playbooks/secrets-init.yml
```diff
@@ -52,13 +52,9 @@
...
           when: vault_init | default('') == 'plugin'
 
-        - name: Populate secrets.yml with missing secrets
-          ansible.builtin.shell: "{{ base_folder }}/scripts/generate-secret.sh -s {{ item }} -m {{ gs_mode }} >> {{ secrets_file }}"
-          changed_when: true
-          when: not lookup('vars', item, default=False)
-          tags:
-            - molecule-idempotence-notest
-          loop:
+        - name: Generate loop of mandatory passwords when empty inventory is provided
```
gionn, line 55: what about dropping this task and `when` on the next one and adding `|| (groups['all'] | length == 0)` to all the new following tasks?

gionn: you can also set a fact `empty_inventory` to avoid repeating it and also make the intention more clear

pmacius: Yeah it'll be better

gionn (approved):

PR #807, opened 2024-03-05
OPSEXP-2551 Fixup redirect url configuration for adw/acc

gionn:
Ref: OPSEXP-2551

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -144,7 +144,7 @@
         "filename": "playbooks/acs.yml",
         "hashed_secret": "3a0b8a438a9efa61267357269709a946d797b9bd",
         "is_verified": false,
-        "line_number": 394,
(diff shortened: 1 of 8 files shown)


alxgomz (commented):
I think we should instead have a generic role param that states the context of the app, and build URLs from the role itself using this param.
What do you think?

alxgomz (changes requested):
Implementation works for me but if I read it correctly, you're not passing the `adf_app_context` so it'll default to `adf_app_name`  which are actually different from the current contexts we have:
alfresco-digital-workspace = > /workspace/
alfresco-control-center => /control-center/

gionn:
> you're not passing the `adf_app_context`

just a bug, now there are tests to properly validate it

alxgomz (approved):

PR #821, opened 2024-04-02
Adopt jekyll for docs website

gionn:
Add workflow and config for publishing docs via jekyll on github pages.

Look and feel customized with the current alfresco docs.


.github/workflows/docs.yml
```diff
@@ -0,0 +1,65 @@
...
+      run:
+        working-directory: docs
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '3.1' # Not needed with a .ruby-version file
```
pmacius, line 32: There is .ruby-version file. Is there a reason to keep this variable?

gionn: lol good catch - even funnier that such comment is coming from the readme of setup-ruby action

pmacius (approved):

.github/workflows/enteprise.yml
```diff
@@ -28,7 +28,7 @@ jobs:
   docker:
     name: Test ${{ matrix.role.name }} role on ${{ matrix.molecule_distro.image }}
     runs-on: ubuntu-latest
-    if: github.repository_owner == 'Alfresco'
+    if: ${{ ! github.event.pull_request.head.repo.fork || github.event_name == 'push' }}
```
alxgomz, line 31: That would fail on forked repo outside of the org (when pushing to forked master branch without a PR) and be skipped in case of PR from a fork within the Alfresco org.

alxgomz (changes requested):
please update action ref to use shasums

.github/workflows/enteprise.yml
gionn: the problem here is just that we don't want to run this workflow at all when secrets are not available

the previous `if: github.repository_owner == 'Alfresco'` was not working either since I switched to `pull_request` (not sure 100% but I've seen it failing on this PR)

docs says we cannot directly reference secrets inside job conditionals, but I have an idea I want to try, in another PR

gionn: https://github.com/Alfresco/alfresco-ansible-deployment/pull/823

alxgomz (commented):
I bothers me a bit we have to change that doc title for table-ish thing. Can we not generate jekyll friendly doc on the fly without impacting the doc people can read on gha?

gionn:
We can't really remove that `---` block which is used to define variables used in jekyll template rendering (https://jekyllrb.com/docs/front-matter/) but we can keep the h1 titles and they even look better on the website

gionn:
Screenshot 2024-04-04 at 12 20 45

alxgomz (approved):

PR #831, opened 2024-04-04
OPSEXP-2610 Move tests for version 7.1 and 7.2 to manually triggered workflow

pmacius:
Ref: OPSEXP-2610


.github/workflows/enteprise.yml
```diff
@@ -173,62 +175,52 @@ jobs:
...
-          retry_wait_seconds: 60
-          command: ./tests/molecule_it/script.sh destroy
+          matrix_name: ${{ matrix.molecule_scenario.name }}
+          matrix_vars: ${{ matrix.molecule_scenario.vars }}
+          matrix_desc: ${{ matrix.molecule_scenario.desc }}
+          nexus_username: ${{ secrets.NEXUS_USERNAME }}
+          nexus_password: ${{ secrets.NEXUS_PASSWORD }}
+          aws_access_key_id: ${{ secrets.aws_access_key_id }}
+          aws_secret_access_key: ${{ secrets.aws_secret_access_key }}
+          BOT_GITHUB_TOKEN: ${{ secrets.BOT_GITHUB_TOKEN }}
```
gionn, line 226: since this action is not really meant to be reusable outside here you can avoid defining those inputs and use `${{ secrets.something }}` directly in the composite action

.github/workflows/enteprise.yml
```diff
@@ -173,62 +175,52 @@ jobs:
...
-      - name: Cleanup tests
-        if: always()
-        timeout-minutes: 65
-        env:
-          MOLECULE_IT_SCENARIO: ${{ matrix.molecule_scenario.name }}
-          MOLECULE_IT_CONFIG: ${{ matrix.molecule_scenario.vars }}
-          AWS_ACCESS_KEY_ID: ${{ secrets.aws_access_key_id }}
-          AWS_SECRET_ACCESS_KEY: ${{ secrets.aws_secret_access_key }}
-        uses: nick-fields/retry@v3
+      - name: Run tests manaully
```
gionn, line 215:
```suggestion
      - name: Run extended tests
```

.github/workflows/enteprise.yml
```diff
@@ -173,62 +175,52 @@ jobs:
...
+          matrix_name: ${{ matrix.molecule_scenario.name }}
+          matrix_vars: ${{ matrix.molecule_scenario.vars }}
+          matrix_desc: ${{ matrix.molecule_scenario.desc }}
+          nexus_username: ${{ secrets.NEXUS_USERNAME }}
+          nexus_password: ${{ secrets.NEXUS_PASSWORD }}
+          aws_access_key_id: ${{ secrets.aws_access_key_id }}
+          aws_secret_access_key: ${{ secrets.aws_secret_access_key }}
+          BOT_GITHUB_TOKEN: ${{ secrets.BOT_GITHUB_TOKEN }}
+
+  ec2-manual:
```
gionn, line 191: don't reference the event name, now that I see it I think it can be an idea to run this "extended tests" on pull_request merge and/or on a weekly schedule and/or applying manually a label on an open PR 🤔

.github/actions/ec2/action.yml
gionn, line 1: add an entry in dependabot.yml to keep this new action updated

.github/workflows/enteprise.yml
pmacius: Actions dont have access to secrets, reusable workflows do.

.github/actions/ec2/action.yml
alxgomz, line 1: Can you maybe rename the workflow and its file? I mean "ec2" made kind of sense as a job name which belonged to a workflow with all other integration tests but if we make it a composite action I think we should be more descriptive (and tbh, we should have been also before...)
something like `molecule_integration_ec2` or the like...

alxgomz, line 1: Please bump Alfresco-build-tools actions to our latest tag and use sha ref for external actions.

alxgomz (changes requested):
What's the reason for choosing the job skipping approach (github.event_name != 'workflow_dispatch')?
A dedicated `on.workflow_dispatch` sounded a lot cleaner to me.

pmacius:
> What's the reason for choosing the job skipping approach (github.event_name != 'workflow_dispatch')? A dedicated `on.workflow_dispatch` sounded a lot cleaner to me.

I am not sure what you mean by dedicated on.workflow_dispatch. In definition of workflow there is workflow_dispatch but we don't want to run every job from this workflow when triggering it. We want to have possibility to run only the job that is responsible for testing older version so we are skipping previous ones if its triggered with workflow dispatch. Also I see same approach on release.yml workflow.

.github/workflows/enteprise.yml
pmacius: https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#accessing-your-secrets

pmacius:
Manual trigger tested on: https://github.com/Alfresco/alfresco-ansible-deployment/actions/runs/8603115779

gionn (approved):

PR #833, opened 2024-04-09
OPSEXP-2549 Set reposearch_shared_secret when search is defined

pmacius:
Ref: OPSEXP-2549

### DIFF
--- a/roles/repository/tasks/main.yml
+++ b/roles/repository/tasks/main.yml
@@ -22,7 +22,7 @@
       ansible.builtin.set_fact:
         temp_array: >-
           {{ temp_array + [ '-Dsolr.sharedSecret=' ~ reposearch_shared_secret | quote ] }}
-      when: (acs.version is version('7.2.0', '>=')) and ((groups.search | default([])) | length > 0)
+      when: (acs.version is version('7.2.0', '>=')) and (reposearch_shared_secret is defined)
 
     - name: Set metadata-keystore.password and related properties
       ansible.builtin.set_fact:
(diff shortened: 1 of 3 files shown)


roles/repository/tasks/main.yml
```diff
@@ -18,11 +18,11 @@
       ansible.builtin.set_fact:
         temp_array: []
 
-    - name: Set solr.sharedSecret property
+    - name: Set solr.sharedSecret property when search is enabled
       ansible.builtin.set_fact:
         temp_array: >-
           {{ temp_array + [ '-Dsolr.sharedSecret=' ~ reposearch_shared_secret | quote ] }}
-      when: acs.version is version('7.2.0', '>=')
+      when: (acs.version is version('7.2.0', '>=')) and ((groups.search | default([])) | length > 0)
```
gionn, line 25: it works, but you can't really have logic related to the inventory inside a role

role should just handle the presence or not of this secret (eventually making it an explicit role argument)

the logic should be in the parent playbook where we have put most of this kind of logic - now I am probably remembering why we created a separate task for this

gionn (changes requested):

roles/repository/tasks/main.yml
```diff
@@ -18,11 +18,11 @@
       ansible.builtin.set_fact:
         temp_array: []
 
-    - name: Set solr.sharedSecret property
+    - name: Set solr.sharedSecret property when search is enabled
```
gionn, line 21:
```suggestion
    - name: Set solr.sharedSecret property when search secret is available
```

gionn (approved):

gionn (approved):

PR #841, opened 2024-04-17
OPSEXP-2627: bump components for Fornax release

alxgomz:
Ref: OPSEXP-2627

Tests of Deprecated versions: https://github.com/Alfresco/alfresco-ansible-deployment/actions/runs/8733936648

### DIFF
--- a/group_vars/all.yml
+++ b/group_vars/all.yml
@@ -182,11 +182,11 @@ use_custom_keystores: false
 dependencies_version:
   postgresql_connector: 42.6.1
   postgres_major_version: 15
-  activemq: 5.18.3
+  activemq: 5.18.4
   java: 17.0.9
   java_build: 9
-  tomcat: 10.1.19
-  libreoffice: 7.2.5.1
+  tomcat: 10.1.20
+  libreoffice: 7.2.5.2
   pdf_renderer: 1.1
   imagemagick: 7.1.0-16-ci-6
 apache_archive_url: https://archive.apache.org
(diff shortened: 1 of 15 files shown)


alxgomz:
> We won't be updating test configs for 7.1 and 7.2?

Well spotted! I missed it since we removed the deprecated versions from the automated CI. Wil l run the job for deprecated versions as it's a release merge

pmacius (approved):

group_vars/all.yml
```diff
@@ -181,11 +181,11 @@ use_custom_keystores: false
 # to default ansible merging behaviour
 dependencies_version:
   postgresql_connector: 42.6.1
-  postgres_major_version: 15
```
gionn, line 184: the last release we did included 15, if now we are reverting this is a breaking change for existing environments

I am not sure how but if it's required only for centos 7 then only centos 7 should stick to 14

gionn (changes requested):

group_vars/all.yml
gionn: the only less ugly thing I can think of, is to add an assert task in the pre-checks playbook that if using postgres and current os is centos 7 and current postgres > 14 then they should downgrade

alxgomz: I'm not sure what use case you are trying to cover for. If users specify a postgres host then it should not have any postgres installation on it - neither from a manual installation, nor from a a run of a previous version of the playbook - as that's not something we support

gionn (approved):

PR #843, opened 2024-04-19
OPSEXP-2627 Update release section with ecosystem updates

gionn:


pmacius (approved):

alxgomz (commented):
I'm not convinced we should do that just with the release PR.

gionn:
> I'm not convinced we should do that just with the release PR.

as a reminder for the worst case, because if we are not actively adding new feature in the playbook we are never gonna merge dependency updates

alxgomz:
> > I'm not convinced we should do that just with the release PR.
> 
> as a reminder for the worst case, because if we are not actively adding new feature in the playbook we are never gonna merge dependency updates

Yes I understand the motivation but bumping libs more or less blindly just before release sounds risky. Maybe we should be more diligent in reviewing dependabot pr BEFORE testing week instead.

gionn:
our test coverage is pretty high here, I've zero concerns merging dependencies if build is green. Most of them are actually just test dependencies and would not impact the end users.

alxgomz (approved):

PR #867, opened 2024-05-21
OPSEXP-2554 Enable using activemq password with $ for trouter role

pmacius:
Ref: OPSEXP-2554


roles/trouter/tasks/main.yml
```diff
@@ -60,21 +55,18 @@
...
+        mode: u=rwx,g=rw,o-rwx
         owner: "{{ username }}"
         group: "{{ group_name }}"
-        marker: "# {mark} ATS COMMON ENV VARS"
-        block: |
-          export ATS_HOME={{ binaries_folder }}/transform-service
-          export ATS_TENGINE_AIO_HOST={{ ats_tengine_aio_host }}
-          export ATS_SHARED_FS_HOST={{ sfs_host }}
-        insertafter: EOF
+      changed_when: false
```
gionn, line 65: try with https://github.com/ansible/ansible/issues/30226#issuecomment-433643224

so we still get changed when the file is created but should not yield changed on the next runs

roles/trouter/templates/alfresco-transform-router.service.j2
gionn, line 1: once both pr are merged we need to think how to get rid this duplication

gionn (changes requested):

roles/trouter/tasks/main.yml
```diff
@@ -60,21 +55,19 @@
...
         - Restart-trouter
 
-    - name: Add paths to setenv file
-      ansible.builtin.blockinfile:
-        path: "{{ config_folder }}/setenv.sh"
+    - name: Create log file with correct permissions
+      ansible.builtin.file:
+        path: "{{ logs_folder }}/ats-atr.log"
+        state: touch
+        mode: u=rwx,g=rw,o-rwx
```
gionn, line 62:
```suggestion
        mode: "0640"
```

gionn (approved):

PR #872, opened 2024-05-28
OPSEXP-2676 Support latest os minor releases

gionn:
Ref: OPSEXP-2676

* Bump Rocky Linux/RHEL to 9.4 and 8.9


pmacius (approved):
Looks good

PR #889, opened 2024-06-25
OPSEXP-2636 Set passphrase secrets when mtls is required and capable

pmacius:
Ref: OPSEXP-2636


playbooks/acs.yml
```diff
@@ -24,6 +24,9 @@
     p12_passphrase: "{{ hostvars.localhost.certs_p12_passphrase }}"
     secret_ca_passphrase: "{{ hostvars.localhost.ca_signing_key_passphrase }}"
   ansible.builtin.import_playbook: pki.yml
+  when:
+    - (groups.repository | difference(groups.transformers | default([]))) | length > 0
```
gionn, line 28: better set a fact on all the hosts rather than duplicating the code

gionn (changes requested):

playbooks/secrets-init.yml
```diff
@@ -55,15 +55,23 @@
...
 
+        - name: Append password_loop when certs_p12_passphrase is needed
+          ansible.builtin.set_fact:
+            password_loop: "{{ password_loop + ['certs_p12_passphrase'] }}"
+          when: (((groups.repository | default([]) | difference(groups.transformers | default([]))) | length > 0) and ats_mtls_capable) or empty_inventory
+
+        - name: Append password_loop when ca_signing_key_passphrase is needed
+          ansible.builtin.set_fact:
+            password_loop: "{{ password_loop + ['ca_signing_key_passphrase'] }}"
+          when: (((groups.repository | default([]) | difference(groups.transformers | default([]))) | length > 0) and ats_mtls_capable) or empty_inventory
```
gionn, line 73: reuse the same var also here no?

gionn (changes requested):

playbooks/mtls-check.yml
```diff
@@ -0,0 +1,18 @@
+---
+- name: Ensure mTLS requirements are met
+  hosts: all:localhost
```
pmacius, line 3:
```suggestion
  hosts: repository:transformers:localhost
```

Running on localhost only was failing, wonder if that would be correct

gionn: of course the fact needs to be set on all the hosts which use it (not only localhost).

One last thing: it's not really required to run `Assert search-enterprise condition` while doing secrets-init, so I would move the task where it was initially, and then rename this new playbook to facts.yml and we will try to use it to populate facts required by most playbooks

gionn (changes requested):

playbooks/mtls-check.yml
pmacius: Decided to change this line to `hosts: localhost:all:!external` due to errors at: https://github.com/Alfresco/alfresco-ansible-deployment/actions/runs/9740313021/job/26880299719#step:3:1043

playbooks/facts.yml
```diff
@@ -0,0 +1,8 @@
+---
+- name: Ensure mTLS requirements are met
```
gionn, line 2:
```suggestion
- name: Populate common facts
```

playbooks/prerun-checks.yml
```diff
@@ -41,15 +41,9 @@
...
-  gather_facts: false
+- name: Assert search-enterprise condition
+  hosts: localhost
   tasks:
-    - name: Check wether we want to enable mTLS for ATS deployment
-      ansible.builtin.set_fact:
-        ats_mtls_required: >-
-          {{ groups.repository | difference(groups.transformers | default([])) | length > 0 }}
-
     - name: Assert that search-enterprise group is empty when ATS mTLS is required
```
gionn, line 47: move this task to the first tasks block (which still runs on localhost)

gionn (changes requested):

gionn (approved):

PR #891, opened 2024-06-27
OPSEXP-2693 Drop support for Centos7

pmacius:
Ref: OPSEXP-2693


gionn (changes requested):
check for centos strings in the entire repo, I see a couple of things which could be dropped too

gionn (approved):

PR #894, opened 2024-07-01
OPSEXP-2637: refactor postgres role to generate syncservice database credentials only when needed

alxgomz:
Ref: OPSEXP-2637


playbooks/secrets-init.yml
pmacius, line 1: Sync password is still in: 
`        - name: Generate loop of mandatory passwords`

pmacius (commented):
I would also change PR title to something more than optional credentials

playbooks/acs.yml
```diff
@@ -51,16 +51,41 @@
...
+        postgres_db_username: "{{ sync_db_username }}"
+        postgres_db_password: "{{ sync_db_password }}"
+        postgres_db_clients: "{{ groups.syncservice }}"
+      when: sync_db_url == "" and groups.syncservice | default([]) | length > 0
   post_tasks:
     - name: Make sure PostgreSQL is running
       ansible.builtin.service:
         name: "{{ postgresql_service }}"
         state: started
+      when: repo_db_url == "" or (sync_db_url == "" and groups.syncservice | default([]) | length > 0)
```
gionn, line 88: set two new vars in this run to avoid duplicating both `repo_db_url == ""` and `sync_db_url == "" and groups.syncservice | default([]) | length > 0`

roles/postgres/meta/argument_specs.yml
```diff
@@ -0,0 +1,28 @@
...
+        type: str
+        required: true
+        description: |
+          Password of the database user
+      postgres_db_clients:
+        type: list
+        elements: str
+        required: true
+        description: |
+          List of clients that are allowed to connect to the database
```
gionn, line 26:
```suggestion
          List of clients addresses in CIDR notation that are allowed to connect to the database
```

?

roles/postgres/tasks/main.yml
```diff
@@ -4,6 +4,10 @@
   loop: >-
     {{ lookup('first_found', os_fallback, errors='ignore', wantlist=True) }}
 
+- name: Set postgreSQL service name as fact
```
gionn, line 7:
```suggestion
- name: Set PostgreSQL service name as fact
```

roles/postgres/tasks/setup_db.yml
```diff
@@ -0,0 +1,63 @@
+---
+- name: Include OS specific variables
+  ansible.builtin.include_vars: "{{ item }}"
+  loop: >-
+    {{ lookup('first_found', os_fallback, errors='ignore', wantlist=True) }}
+
+- name: Configure postgresql client auth
```
gionn, line 7:
```suggestion
- name: Configure PostgreSQL client auth
```

roles/postgres/tasks/setup_db.yml
```diff
@@ -0,0 +1,63 @@
...
+  become: true
+  notify:
+    - Restart-postgresql
+  ansible.builtin.blockinfile:
+    path: "{{ postgresql_conf_path }}/pg_hba.conf"
+    block: |
+      {% for host in postgres_db_clients | map('extract', hostvars, ['ansible_default_ipv4', 'address']) %}
+      host {{ postgres_db_name }} {{ postgres_db_username }} {{ host }}/32 md5
```
gionn, line 15: ok if we want to enforce the /32 here maybe better to make it clear in the arguments description

roles/postgres/tasks/setup_db.yml
```diff
@@ -0,0 +1,63 @@
...
+      {% endfor %}
+    marker: >-
+      # {mark} ANSIBLE MANAGED: allow {{ postgres_db_clients | join(", ") }} to connect to {{ postgres_db_name }} as {{ postgres_db_username }}
```
gionn, line 18: is it ok that the marker contains variables? isn't gonna mess updates if those variables changes on a next run?

roles/postgres/tasks/setup_db.yml
```diff
@@ -0,0 +1,63 @@
...
+    owner: postgres
+    group: postgres
+    mode: "u=rw"
+
+- name: Configure PostgreSQL database
+  become: true
+  become_user: postgres
+  vars:
+    ansible_ssh_pipelining: true
```
gionn, line 27: should not be declared here

roles/postgres/tasks/setup_db.yml
```diff
@@ -0,0 +1,63 @@
...
+  block:
+    - name: Create necessary databases
```
gionn, line 29:
```suggestion
    - name: Create database
```

roles/postgres/tasks/setup_db.yml
```diff
@@ -0,0 +1,63 @@
...
+        db: "{{ postgres_db_name }}"
+        privs: ALL
+        type: schema
+        objs: public
+        role: public
+        state: absent
+      tags:
+        - molecule-idempotence-notest
+
+    - name: Create unprivileged users
```
gionn, line 44:
```suggestion
    - name: Create unprivileged user
```

roles/postgres/tasks/setup_db.yml
```diff
@@ -0,0 +1,63 @@
...
+        db: "{{ postgres_db_name }}"
+        name: "{{ postgres_db_username }}"
+        password: "{{ postgres_db_password }}"
+        expires: infinity
+        role_attr_flags: NOSUPERUSER
+      no_log: true
+      tags:
+        - molecule-idempotence-notest
+
+    - name: Grant db privileges to users
```
gionn, line 55:
```suggestion
    - name: Grant db privileges to user
```

roles/sync/molecule/default/converge.yml
```diff
@@ -2,12 +2,40 @@
...
-    - role: repository
-      repository_properties: "{{ global_properties }}"
     - role: nginx
-    - role: sync
+  tasks:
+    - name: Install PostgreSQL
+      ansible.builtin.include_role:
+        name: postgres
+    - name: Flush handlers
+      ansible.builtin.meta: flush_handlers
```
gionn, line 12: it's already at the end of the main postgres task

gionn (changes requested):

roles/postgres/meta/argument_specs.yml
alxgomz: nope, it's actually expecting the inventory hostname list (e.g. the inventory group(s))=

roles/postgres/tasks/setup_db.yml
alxgomz: I'm not sure what behaviour we'd really want in such case. With current code, changing the clients, username or db name would result in additional entries while with no variable in would replace the old one with the new one but would also prevent passing multiple databases unless we keep at least the db name in the marker. Then still, it would not be possible to pass multiple users/clients entries for the same db....
All in all I think the current approach is better than other that bring these limitations.

roles/postgres/tasks/setup_db.yml
alxgomz: It is purposely enforced here to ensure the task would not fail even is user changes the ansible global  configuration to disable pipelining.
See https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_privilege_escalation.html#risks-of-becoming-an-unprivileged-user

roles/postgres/tasks/setup_db.yml
alxgomz: the argument desc already mentions a list of inventory hosts is expected so it would always be /32

playbooks/secrets.yml
pmacius, line 1: Should there be also a task like `name: Ensure activemq_password is set when required` for sync_db_password password?

alxgomz: Of course! 👍

pmacius (approved):

roles/postgres/tasks/setup_db.yml
gionn: from the docs I am not understanding why pipelining should be enabled when becoming an unprivileged user

roles/postgres/tasks/setup_db.yml
gionn: not following you, I was looking that you used variables inside a marker, but looking at the docs looks like that `{mark}` is the only thing really used by ansible to recognize the block

gionn (approved):
check the last 2 open requests but should be ok

roles/postgres/tasks/setup_db.yml
alxgomz: to avoid permission issues when dealing with temp module files

roles/postgres/tasks/setup_db.yml
alxgomz: `{mark}` evaluate either to `BEGIN` or `END` depending on whether it marks the beginning or the end of the block.

So if you call the `setup_db` endpoint twice with 2 different dbs or/and hosts or/and users and/or clients, but the mark is the same for both iterations, the second block will replace the first one

PR #901, opened 2024-07-09
OPSEXP-2739: bump Alfresco components and third parties

alxgomz:
Ref: OPSEXP-2739


.github/workflows/precommit.yml
```diff
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - master
+  workflow_dispatch:
```
pmacius, line 8: Do we really need this? What are the use cases?

alxgomz: This is for when the last commit has a `[skip ci]` flag set so the workflow can be manually triggered (as it is required for PR merge)

docs/overview.md
```diff
@@ -76,24 +76,24 @@ distributions.
...
-| Apache Tomcat       | 10.1.20                     | 9.0.86         | 9.0.86         | 9.0.86         | 9.0.86         |
+| OpenJDK             | 17.0.11                     | 17.0.11        | 11.0.23        | 11.0.23        | 11.0.23        |
+| Apache Tomcat       | 10.1.25                     | 9.0.91         | 9.0.91         | 9.0.91         | 9.0.91         |
 | PostgreSQL          | 15.x                        | 14.x           | 14.x           | 13.x           | 13.x           |
 | Apache ActiveMQ     | 5.18.4                      | 5.17.6         | 5.17.6         | 5.16.7         | 5.16.7         |
-| Repository          | 23.2.1                      | 7.4.2          | 7.3.2          | 7.2.2          | 7.1.1.10       |
-| Share               | 23.2.1                      | 7.4.2          | 7.3.2          | 7.2.2          | 7.1.1.10       |
-| Search Services     | 2.0.10                      | 2.0.10         | 2.0.10         | 2.0.10         | 2.0.2.2        |
-| Search Enterprise   | 4.0.1 (Enterprise only)     | 3.3.3          | 3.2.1          | 3.1.1.1        | 3.1.1.1        |
```
gionn, line 88: I liked the explicit `(Enterprise only)`, `(n/a)` doesn't help at all if you didn't notice also the column heading

molecule/default/vars-rocky9.yml
```diff
@@ -1,4 +1,4 @@
-MOLECULE_IT_IMAGE_ID: ami-0d071a78283fb19c3 # Rocky-9-EC2-Base-9.4-20240509.0.x86_64
+MOLECULE_IT_IMAGE_ID: ami-0230bf6b41b114fef # Rocky-9-EC2-Base-9.4-20240509.0.x86_64
```
gionn, line 1: didn't update also the comment

molecule/default/vars-ubuntu18.yml
```diff
@@ -1,4 +1,4 @@
-MOLECULE_IT_IMAGE_ID: ami-08bd6fc5c8daf3309 # ubuntu/images/hvm-ssd/ubuntu-bionic-18.04-amd64-server-20230517
+MOLECULE_IT_IMAGE_ID: ami-02076d50a3b3930d0 # ubuntu/images/hvm-ssd/ubuntu-bionic-18.04-amd64-server
```
gionn, line 1: timestamp was cool to see at glance how old it is

docs/overview.md
alxgomz: I understand but keeping the previous heading would end up with an odd column containing 2 versions separated with just a space

molecule/default/vars-ubuntu18.yml
alxgomz: same as above

gionn (approved):

PR #937, opened 2024-09-12
OPSEXP-2827: cover for system greedy expansions

alxgomz:
Ref: OPSEXP-2827

%z is expanded although not a specifier


pmacius (commented):
Not sure about this. If there will be `%%` in string it will be changed `%`. Maybe if we just replace any % sign with double percent it will be good. We don't have to specify any char after that it will be always replaced so we don't have to worry. If I understand that correctly of course.
"%%" | Single percent sign | Use "%%" in place of "%" to specify a single percent sign.
-- | -- | --

PR #947, opened 2024-09-23
OPSEXP-2121 Add kics scan

pmacius:
Ref: OPSEXP-2121


.github/workflows/kics.yml
```diff
@@ -0,0 +1,40 @@
...
+permissions:
+  pull-requests: write
+  contents: write
+
+jobs:
+  kics:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - name: run kics Scan
```
gionn, line 28:
```suggestion
      - name: Run KICS Scan
```

.github/workflows/kics.yml
```diff
@@ -0,0 +1,40 @@
...
+        uses: checkmarx/kics-github-action@530ac1f8efe6202b0f12c9a6e952597ae707b755 # v2.1.2
+        with:
+          path: 'playbooks,roles'
+          ignore_on_exit: results
+          output_path: report-dir/
+          output_formats: 'sarif'
+          token: ${{ secrets.GITHUB_TOKEN }}
+          enable_jobs_summary: true
```
gionn, line 36:
```suggestion
          enable_jobs_summary: true
          enable_comments: false
```

we agreed on that right?

.github/workflows/kics.yml
```diff
@@ -0,0 +1,40 @@
...
+
+jobs:
+  kics:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      - name: run kics Scan
+        uses: checkmarx/kics-github-action@530ac1f8efe6202b0f12c9a6e952597ae707b755 # v2.1.2
+        with:
+          path: 'playbooks,roles'
```
gionn, line 31: maybe also add `platform_type: ansible` if make the scan faster?

gionn (changes requested):
we could also disable secret scanning with `disable_secrets` given that we are already using detect-secrets in pre-commit

.github/workflows/kics.yml
```diff
@@ -0,0 +1,40 @@
+name: kics
+
+on:
+  pull_request:
+    branches: [master]
+    paths-ignore:
+      - "docs/**"
+      - "*.md"
+      - .github/workflows/docs.yml
+      - .pre-commit-config.yaml
```
gionn, line 10: given you have `path: 'playbooks,roles'` I would add the same as paths here

.github/workflows/kics.yml
pmacius: enable_comments defaults to false

gionn: https://github.com/Checkmarx/kics-github-action?tab=readme-ov-file#inputs

I was assuming the first column was `default` 🤦🏻

.github/workflows/kics.yml
```diff
@@ -0,0 +1,41 @@
...
+      - 'playbooks/**'
+      - 'roles/**'
+      - '.github/workflows/kics.yml'
+  push:
+    branches: [master]
+    paths-ignore:
+      - "docs/**"
+      - "*.md"
+      - .github/workflows/docs.yml
+      - .pre-commit-config.yaml
```
gionn, line 16: same as before for consistency

.github/workflows/kics.yml
pmacius: Yeah not very intuitive - also the column with default values is hidden a little :D

.github/workflows/kics.yml
```diff
@@ -0,0 +1,41 @@
...
+
+permissions:
+  pull-requests: write
+  contents: write
```
gionn, line 20: are those still required with comments disabled?

pmacius: Ahh right let me test not sure - probably contents still needs to be there to upload sarif.

pmacius: not permission required

gionn (approved):

PR #996, opened 2024-11-26
OPSEXP-2880 Add audit-storage role

gionn:
OPSEXP-2880


pmacius (approved):
Are we disabling this role for older version?

gionn:
> Are we disabling this role for older version?

good catch, let me add an explicit validation for that

pmacius (commented):
We can bump dtas version to v1.6.0 and configure tests for audit-storage

pmacius (dismissed):
LGTM

roles/audit_storage/defaults/main.yml
```diff
@@ -0,0 +1,48 @@
...
+audit_storage_artifact_name: alfresco-audit-storage-app
+
+audit_storage_username: alfresco
+audit_storage_group_name: alfresco
+
+audit_storage_server_port: 8083
+
+audit_storage_broker_url: failover:(nio://localhost:61616)?timeout=3000
+audit_storage_broker_username: admin
+audit_storage_broker_password: admin
```
alxgomz, line 16: why these default when we just don't set passwd for ES?

alxgomz (commented):
Just a few comments, but maybe fix the password defaults in the module if there's not reason for them?

alxgomz (approved):

PR #1006, opened 2024-12-03
OPSEXP-2941 Bump readme and stack patches

gionn:
* Make sure that activemq version is properly propagated in downstream usage
* Build audit-storage java path from java role var

OPSEXP-2941


pmacius (dismissed):

README.md
```diff
@@ -192,31 +191,20 @@ doc](./README.md#versioning)
...
+1. After merging every pending PR, proceed with tagging:
+   * `git tag -s v2.x.x -m v2.x.x`
+   * `git push origin v2.x.x`
+1. Wait for the [Release
+  workflow](https://github.com/Alfresco/alfresco-ansible-deployment/actions/workflows/release.yml)
+  go green.
+1. [Draft a new
+  release](https://github.com/Alfresco/alfresco-ansible-deployment/releases) on
+  GitHub with the tag you just pushed. If the release is for a new ACS major
+  version, mention the ACS release in the title, e.g. v2.x.x (ACS 23.4.0)
```
alxgomz, line 210: Does incrementing the number breaks the rendering?  If not then please do increment. If it does... well it shouldn't :)

roles/activemq/defaults/main.yml
```diff
@@ -9,4 +9,4 @@ activemq_environment:
     - -Xms128m
     - -Xmx1g
     - $ACTIVEMQ_OPTS
-activemq_version: 5.18.5
+activemq_version: null
```
alxgomz, line 12: hat's the benefit vs removing it from here?

alxgomz (changes requested):

roles/activemq/defaults/main.yml
gionn: the role in the current state is not really reusable (the download URL is not using this activemq_version) - so I prefer to get an explicit failure when we forget to pass this variable properly

README.md
gionn: may look confusing at first but that's the way you get numbers automatically managed in markdown - I think it was ok given users are supposed to browse the jekyll website to read docs

https://www.markdownguide.org/basic-syntax/#ordered-lists

PR #1008, opened 2024-12-05
Fixup base folder in apply_amps

### LINKED ISSUE #995: Wrong modules path in apply_amps.sh
Since 7.1.1 the modules/acs-platform and modules/acs-share moved from /opt/alfresco to /opt/alfresco/content-services-<acs-version>. This changeaffects the apply-amps.sh file that still references the modules root path being `{{ binaries_folder }}` while it should not use the already existing ${ACS_HOME} variable.

https://github.com/Alfresco/alfresco-ansible-deployment/blob/8883dfb305a5c22a62610197797d25acb8946655/roles/repository/templates/apply_amps.sh#L5C74-L5C75

gionn:
Fix #995


roles/repository/templates/apply_amps.sh.j2
alxgomz, line 1: As the command task runs with a `become: true` we should maybe remove the `sudo` which is not even documented as a requirement

pmacius (approved):

PR #1009, opened 2024-12-05
OPSEXP-2930 Point links to support.hyland.com

pmacius:
Ref: OPSEXP-2930

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -6,6 +6,7 @@ color_scheme: alfresco
 url: https://alfresco.github.io
 
 aux_links:
+  Hyland Support: https://support.hyland.com/p/alfresco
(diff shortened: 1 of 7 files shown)


docs/_config.yml
```diff
@@ -6,6 +6,7 @@ color_scheme: alfresco
 url: https://alfresco.github.io
 
 aux_links:
+  Hyland Support: https://support.hyland.com/p/alfresco
   Docs: https://docs.alfresco.com/
```
gionn, line 10: should we remove this since it's deprecated or rename it to legacy docs?

pmacius: I would rename it. 7.3 and 7.4 are still supported but not covered in new docs

gionn (approved):

PR #1018, opened 2025-01-09
OPSEXP-2955 Avoid resetting permissions on mounted shared content storage

gionn:
OPSEXP-2955


pmacius (approved):
Looks good but lets wait for @alxgomz

alxgomz (approved):

PR #1019, opened 2025-01-10
OPSEXP-2948 Prepare for v25 release

gionn:
OPSEXP-2948


roles/sync/molecule/default/verify.yml
```diff
@@ -0,0 +1,56 @@
...
+  tasks:
+    - name: Assert service is running
+      ansible.builtin.systemd:
+        name: "alfresco-sync"
+        state: started
+
+    - name: Assert service is enabled
+      ansible.builtin.systemd:
+        name: "alfresco-sync"
+        enabled: yes
```
alxgomz, line 13: Does molecule verify fail on "changed" task status (like idempotence)? I'm not sure... If not, then here you're running the tasks, not asserting. verify would not fail if service is not started it would just start it.

alxgomz (approved):

PR #1020, opened 2025-01-13
OPSEXP-2954 Support ubuntu 24.04

gionn:
OPSEXP-2954

### DIFF
--- a/.github/workflows/community.yml
+++ b/.github/workflows/community.yml
@@ -52,6 +52,7 @@ jobs:
       fail-fast: false
       matrix:
         molecule_distro:
+          - image: ubuntu:24.04
           - image: ubuntu:22.04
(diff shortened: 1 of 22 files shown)


pmacius (commented):
Cosmetic stuff but we can update `.ansible-lint-ignore` file. I also see 
```
roles/postgres/vars/Ubuntu20.yml var-naming[no-role-prefix]
roles/postgres/vars/Ubuntu22.yml var-naming[no-role-prefix]
```
there but in postgres role we have only ubuntu18 vars

gionn:
> but we can update .ansible-lint-ignore file

see https://github.com/Alfresco/alfresco-ansible-deployment/pull/1026 🪄

alxgomz (approved):

PR #1021, opened 2025-01-13
OPSEXP-2951 Enable elasticsearch 8 major version

pmacius:
Ref: OPSEXP-2951

### DIFF
--- a/roles/audit_storage/molecule/default/converge.yml
+++ b/roles/audit_storage/molecule/default/converge.yml
@@ -6,4 +6,6 @@
       vars:
         activemq_version: "{{ dependencies_version.activemq }}"
     - role: elasticsearch
+      vars:
+        elasticsearch_version: "{{ dependencies_version.elasticsearch_major_version }}"
     - role: audit_storage
--- a/roles/audit_storage/molecule/default/converge.yml
+++ b/roles/audit_storage/molecule/default/converge.yml
@@ -6,6 +6,4 @@
       vars:
         activemq_version: "{{ dependencies_version.activemq }}"
     - role: elasticsearch
-      vars:
-        elasticsearch_version: "{{ dependencies_version.elasticsearch_major_version }}"
     - role: audit_storage
--- a/roles/audit_storage/molecule/default/converge.yml
+++ b/roles/audit_storage/molecule/default/converge.yml
@@ -6,4 +6,6 @@
       vars:
         activemq_version: "{{ dependencies_version.activemq }}"
     - role: elasticsearch
+      vars:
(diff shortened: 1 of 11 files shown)


roles/elasticsearch/tasks/main.yml
```diff
@@ -28,9 +28,12 @@
...
       vars:
+        elasticsearch_version: "{{ elasticsearch_major_version }}"
         elasticsearch_network_host: "{{ ansible_hostname }}"
         elasticsearch_heap_size_min: 1g
         elasticsearch_heap_size_max: 1g
         elasticsearch_extra_options: |
           xpack.security.enabled: false
           discovery.type: single-node
+          xpack.security.transport.ssl.enabled: false
+          xpack.security.http.ssl.enabled: false
```
pmacius, line 39: Testing locally with elasticsearch role molecule required those two settings to work. Works both on 7.x and 8.x

roles/elasticsearch/vars/main.yml
```diff
@@ -1,2 +1,3 @@
 ---
 # vars file for elasticsearch
+elasticsearch_major_version: "{{ '8.x' if acs.version is version('25.0', 'ge') else '7.x' }}"
```
alxgomz, line 3: the elasticsearch role should not know anything about acs. Please move that logic to the playbook

alxgomz (changes requested):

roles/elasticsearch/defaults/main.yml
```diff
@@ -2,3 +2,5 @@
 # defaults file for elasticsearch
 elasticsearch_tmp_path: /usr/share/elasticsearch/tmp
 elasticsearch_systemd_service_path: /etc/systemd/system/elasticsearch.service.d/
+
+elasticsearch_major_version: null
```
gionn, line 6: I would add 8.x as default here, so it serve also as documentation

roles/audit_storage/molecule/default/converge.yml
```diff
@@ -6,4 +6,6 @@
       vars:
         activemq_version: "{{ dependencies_version.activemq }}"
     - role: elasticsearch
+      vars:
+        elasticsearch_major_version: "{{ '8.x' if acs.version is version('25.0', 'ge') else '7.x' }}"
```
gionn, line 10: not sure if it's really worth having this logic for molecule, but it would prevent us to forget switching to 8.x by default, so maybe we can still drop it later

gionn (approved):

alxgomz (approved):

PR #1041, opened 2025-01-28
OPSEXP-3006 Move upgrade checks as playbook and add skip variable

gionn:
* Add an easy way to skip the upgrade checks which can be used to test if an upgrade is really supported

OPSEXP-3006


alxgomz (approved):
Looks like it's lacking a bunch of components checks but let's not add them now

playbooks/prerun-upgrade-checks.yml
```diff
@@ -0,0 +1,78 @@
...
+          ansible.builtin.fail:
+          when:
+            - installed_components.search is defined
+            - search.version != installed_components.search.version
+
+        - name: Fail on unsupported Sync upgrades
+          ansible.builtin.fail:
+          when:
+            - installed_components.sync is defined
+            - sync.version != installed_components.sync.version
```
pmacius, line 78: Can be defined with loop

gionn: it cloud but for sure it will make harder to understand the logs, since you will have only a generic task name and a failure on item=component_name

pmacius (approved):

PR #501, opened 2022-11-14
workaround for user admin in community

alxgomz:
Workaround for  https://github.com/Alfresco/acs-community-packaging/issues/367

PR #1056, opened 2025-02-10
fix case issue on showAuthorizationStatus

alxgomz:
looks like #501 is still not fixed


gionn (approved):
LGTM if manually tested

PR #1059, opened 2025-02-11
OPSEXP-3041 Refactor activemq role to use default vars

pmacius:
Ref: OPSEXP-3041


roles/activemq/meta/argument_specs.yml
```diff
@@ -5,5 +5,11 @@ argument_specs:
...
       activemq_version:
-        description: activemq version to use for deployment (URL still handled via group_vars)
+        description: activemq version to use for deployment
         required: true
+      activemq_dependency_url:
+        description: activemq dependency URL to use for deployment
+        required: false
+      activemq_dependency_sha512_checksum_url:
+        description: activemq dependency sha512 checksum URL to use for deployment
+        required: false
```
alxgomz, line 15: I'd avoid reusing the playbook var names and use a more generic one within the role: `activemq_url` & `activemq_checksum`.
From the role point of view activemq is not a dependency and the checksum could be any other type of hash

roles/activemq/defaults/main.yml
```diff
@@ -9,4 +9,6 @@ activemq_environment:
     - -Xms128m
     - -Xmx1g
     - $ACTIVEMQ_OPTS
-activemq_version: null
+activemq_version: 5.18.6
+activemq_dependency_url: https://archive.apache.org/dist/activemq/{{ activemq_version }}/apache-activemq-{{ activemq_version }}-bin.tar.gz
+activemq_dependency_sha512_checksum_url: https://archive.apache.org/dist/activemq/{{ activemq_version }}/apache-activemq-{{ activemq_version }}-bin.tar.gz.sha512
```
alxgomz, line 14: I would move this variable to the [get_url format](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/get_url_module.html#parameter-checksum) so it can easily be changed

alxgomz (changes requested):

roles/activemq/tasks/main.yml
```diff
@@ -13,14 +13,9 @@
...
 
-- name: Download ActiveMQ archive # noqa ignore-errors
-  delegate_to: localhost
-  ansible.builtin.get_url:
-    url: "{{ dependencies_url.activemq }}"
-    checksum: sha512:{{ dependencies_url.activemq_sha512_checksum_url }}
-    dest: "{{ download_path.path }}/apache-activemq-{{ activemq_version }}-bin.tar.gz"
-    mode: "0644"
-  ignore_errors: true
+- name: Download ActiveMQ {{ activemq_version }} archive from {{ activemq_url }}
```
pmacius, line 16:
```suggestion
- name: Download ActiveMQ archive 
```
Not needed - but I think there is a possible value

playbooks/group_vars/activemq.yml
```diff
@@ -0,0 +1,5 @@
+activemq_v: "{{ acs_play_activemq_version }}"
```
gionn, line 1: why don't use `acs_play_activemq_version` straight everywhere you are using now `activemq_v`?

playbooks/group_vars/activemq.yml
```diff
@@ -0,0 +1,5 @@
...
+apache_archive_url: https://archive.apache.org
+activemq_archive_url: "{{ apache_archive_url }}/dist/activemq"
+activemq_archive_download_url: "{{ activemq_archive_url }}/{{ activemq_v }}/apache-activemq-{{ activemq_v }}-bin.tar.gz"
+activemq_checksum_sha512: "sha512:{{ activemq_archive_download_url }}.sha512"
```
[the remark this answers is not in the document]
pmacius: For me its fine if we rely on default. @alxgomz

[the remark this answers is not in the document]
alxgomz: we can rely on default indeed. I guess they should be documented in the role doc later on

alxgomz (dismissed):

gionn (approved):

PR #1065, opened 2025-02-13
OPSEXP-3043 Refactor audit_storage role to use default vars

pmacius:
Ref: OPSEXP-3043

### DIFF
--- a/playbooks/include-vars.yml
+++ b/playbooks/include-vars.yml
@@ -5,7 +5,7 @@
   vars:
     base_folder: "{{ playbook_dir }}/.."
     vars_folder: "{{ base_folder }}/vars"
-    vars_file: "{{ vars_folder }}/acs{{ acs_major_version }}.yml"
+    vars_file: "{{ vars_folder }}/acs{{ acs_play_major_version }}.yml"
   tasks:
     - name: Include vars based on acs major version
       ansible.builtin.include_vars:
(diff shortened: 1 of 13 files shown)


playbooks/include-vars.yml
gionn, line 1: when you end up in this situation is better that you raise a new PR with just the changes needed in upcoming PRs so you can merge it first more easily

gionn (dismissed):

roles/audit_storage/defaults/main.yml
```diff
@@ -4,7 +4,7 @@ audit_storage_version: "1.0.0"
 audit_storage_zip_url: https://nexus.alfresco.com/nexus/repository/enterprise-releases/org/alfresco/alfresco-audit-storage-distribution/{{ audit_storage_version }}/alfresco-audit-storage-distribution-{{ audit_storage_version }}.zip
 audit_storage_zip_sha1_url: https://nexus.alfresco.com/nexus/repository/enterprise-releases/org/alfresco/alfresco-audit-storage-distribution/{{ audit_storage_version }}/alfresco-audit-storage-distribution-{{ audit_storage_version }}.zip.sha1
```
alxgomz, line 5: please align with checksum format type:hash and rename so the var doesn't contain the hash type

alxgomz (changes requested):

alxgomz (approved):

gionn (approved):

PR #1061, opened 2025-02-11
OPSEXP-3042: set ADF download vars as adf_ap role's arguments

alxgomz:
Ref: OPSEXP-3042

PR #1069, opened 2025-02-17
OPSEXP-3045 Refactor search role to use default vars

pmacius:
ref: OPSEXP-3045


playbooks/group_vars/search.yml
```diff
@@ -0,0 +1,4 @@
+search_artifact_name: alfresco-search-services
+search_repository: "{{ nexus_repository.releases }}"
```
gionn, line 2: no acs_play for these?

pmacius: Followed from https://github.com/Alfresco/alfresco-ansible-deployment/pull/1061/files#diff-d2c6efd9c11af29f1c3a4f419093398b6be55a5cbb93a79696cce166661e9cb7

playbooks/group_vars/search.yml
```diff
@@ -0,0 +1,4 @@
+acs_play_search_artifact_name: alfresco-search-services
+acs_play_search_repository: "{{ nexus_repository.releases }}"
```
alxgomz, line 2: I would not prefix these as they are not role arguments and are only used to build up the artifact url.

roles/search/meta/argument_specs.yml
```diff
@@ -83,13 +83,23 @@ argument_specs:
...
         description: |
           Choose between alfresco-search-services and alfresco-insight-engine.
           Insight Engine requires a specific entitlement.
         required: false
         default: alfresco-search-services
+      search_zip_url:
+        description: |
+          The URL of the Search Service zip file to download.
+        required: false
+        type: str
```
alxgomz, line 97: Wondering if we should add the defaults i here. More painful to maintain but provides more info to users

```suggestion
        type: str
        default: https://artifacts.alfresco.com/nexus/content/groups/public/org/alfresco/2.0.14/alfresco-search-services/2.0.14/alfresco-search-services-2.0.14.zip
```

roles/search/meta/argument_specs.yml
```diff
@@ -83,13 +83,23 @@ argument_specs:
...
+      search_zip_checksum_url:
+        description: |
+          The URL of the Search Service zip file checksum to download.
+        required: false
+        type: str
```
alxgomz, line 102: Wondering if we should add the defaults i here. More painful to maintain but provides more info to users

```suggestion
        type: str
        default: sha1:https://artifacts.alfresco.com/nexus/content/groups/public/org/alfresco/2.0.14/alfresco-search-services/2.0.14/alfresco-search-services-2.0.14.zip.sha1
```

playbooks/group_vars/search.yml
gionn: using here the same prefix of a role prefix is asking for troubles

gionn (approved):

alxgomz (approved):

PR #1071, opened 2025-02-17
OPSEXP-3047 Refactor sfs role to use default vars

pmacius:
ref: OPSEXP-3047

### DIFF
--- a/playbooks/group_vars/all.yml
+++ b/playbooks/group_vars/all.yml
@@ -1,9 +1,2 @@
 default_java_version: 17.0.14+7
 acs_play_java_core: "{{ default_java_version.split('+')[0] }}"
-ats_mtls_capable: >
-  {{
-    acs.version is version('7.4', 'ge') and
-    transform.version is version('3.1', 'ge') and
-    trouter.version is version('2.1', 'ge') and
-    acs_play_sfs_version is version('2.1', 'ge')
-  }}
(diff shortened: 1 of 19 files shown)


playbooks/group_vars/all.yml
```diff
@@ -0,0 +1,7 @@
+ats_mtls_capable: >
+  {{
+    acs.version is version('7.4', 'ge') and
+    transform.version is version('3.1', 'ge') and
+    trouter.version is version('2.1', 'ge') and
+    acs_play_sfs_version is version('2.1', 'ge')
+  }}
```
gionn, line 7: this is entirely based on current versions, so why don't put ats_mtls_capable=true for 7.4 and greater and false for 7.3 in the acs specific vars?

pmacius:
Before rebase: https://github.com/Alfresco/alfresco-ansible-deployment/actions/runs/13409693062

gionn (dismissed):

vars/acs73.yml
```diff
@@ -8,8 +8,10 @@ supported_os:
   Ubuntu:
     versions:
       - 22.04
+ats_mtls_capable: true
```
alxgomz, line 11: how is ats mtls capable in 73 and not in 74?

pmacius: fixing now 💯

alxgomz (approved):

PR #1079, opened 2025-02-18
OPSEXP-3048: add sync download vars as role defaults

alxgomz:
Ref: OPSEXP-3048

### DIFF
--- a/roles/sync/defaults/main.yml
+++ b/roles/sync/defaults/main.yml
@@ -1,4 +1,11 @@
 ---
+sync_version: 5.1.0
+sync_zip_url: >-
+  https://artifacts.alfresco.com/nexus/content/groups/private/org/alfresco/services/sync/sync-dist-6.x/{{ sync_version }}/sync-dist-6.x-{{ sync_version }}.zip
+sync_zip_checksum: >-
+  sha1:{{ sync_zip_url }}.sha1
+sync_zip_username: null
+sync_zip_password: null
 sync_home: "{{ binaries_folder }}/sync-service"
 utils_sync: []
 sync_environment:
--- a/roles/sync/meta/argument_specs.yml
+++ b/roles/sync/meta/argument_specs.yml
@@ -1,0 +1,38 @@
+---
+argument_specs:
+  main:
+    short_description: deploys Alfresco Sync Service
+    description: |
+      This role deploys Alfresco Sync Service. A service which allows syncing
+      content rom the Alfresco repository and Alfresco Desktop Sync (a client
+      application).
+    author: Alfresco
+    options:
+      sync_version:
+        description: |
+          Version of the Alfresco Sync Service to deploy.
+        type: str
+        default: 5.1.0
+      sync_zip_url:
+        description: |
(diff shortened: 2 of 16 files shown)


roles/sync/tasks/main.yml
```diff
@@ -7,19 +7,19 @@
...
-    url: "{{ downloads.sync_zip_url }}"
-    dest: "{{ download_location }}/sync-dist-6.x-{{ sync.version }}.zip"
-    checksum: sha1:{{ lookup('url', downloads.sync_zip_sha1_checksum_url, username=nexus_user, password=nexus_password) }}
-    url_username: "{{ nexus_user }}"
-    url_password: "{{ nexus_password }}"
+    url: "{{ sync_zip_url }}"
+    dest: "{{ download_location }}/sync-dist-6.x-{{ sync_version }}.zip"
+    checksum: "{{ sync_zip_checksum }}"
+    url_username: "{{ sync_zip_username }}"
+    url_password: "{{ sync_zip_password }}"
```
pmacius, line 14: What is the reason to explicitly provide those for sync

roles/sync/defaults/main.yml
```diff
@@ -1,4 +1,11 @@
 ---
+sync_version: 5.1.0
+sync_zip_url: >-
+  https://artifacts.alfresco.com/nexus/content/groups/private/org/alfresco/services/sync/sync-dist-6.x/{{ sync_version }}/sync-dist-6.x-{{ sync_version }}.zip
+sync_zip_checksum: >-
+  sha1:{{ sync_zip_url }}.sha1
+sync_zip_username: null
+sync_zip_password: null
```
pmacius, line 8: Maybe we can default to lookup here like we do for common role

alxgomz: looking up the credentials from the env var is specific to our playbook. One should be able to use the role from a custom play and get the credentials from some other place. So I chose to not default to anything...
Now thinking about it again if I want to be consistent I should make these vars mandatory

pmacius: Ok we should align that behavior in every other role.

roles/sync/tasks/main.yml
gionn: default(omit)

roles/sync/defaults/main.yml
gionn: mandatory may be a strong assumption, and in the end getting a 401 is easy to understand

pmacius (approved):

roles/sync/meta/argument_specs.yml
```diff
@@ -0,0 +1,38 @@
...
+        type: str
+        default: null
+      sync_zip_password:
+        description: |
+          Password to use to download the Alfresco Sync Service zip file.
+        type: str
+        default: null
+      sync_zip_checksum:
+        description: |
+          Checksum of the Alfresco Sync Service zip file folowing the format
```
gionn, line 34:
```suggestion
          Checksum of the Alfresco Sync Service zip file following the format
```

gionn (approved):

PR #1082, opened 2025-02-19
OPSEXP-3052 Refactor trouter role to use default vars

pmacius:
ref: OPSEXP-3052


gionn (dismissed):

alxgomz (commented):
So in the end do we stick to `trouter_zip_*`variable names even for jars?

pmacius:
@alxgomz I was thinking to change the vars to something more consistent for all roles after we end the current loop of PRs

PR #1083, opened 2025-02-19
OPSEXP-3044: pass java_version as an argument for roles depending on java

alxgomz:
Ref: OPSEXP-3044


pmacius (approved):
For java taks that should be enough but I was thinking that we probably should move everything to import_role to be consistent. Check that error on opensearch scenario not sure if thats some flaky test

gionn (dismissed):

pmacius (approved):

PR #1086, opened 2025-02-20
OPSEXP-3053 Refactor transformers_imagemagick to use default vars

pmacius:
ref: OPSEXP-3053


alxgomz (changes requested):
Do you mind changing the `_te_img_` prefix to  `_te_im_` as this is what we use in other deployment tools?

gionn (approved):

PR #1088, opened 2025-02-21
OPSEXP-3055 Refactor repository war downloads to use defaults

pmacius:
ref: OPSEXP-3055


gionn (commented):
I am not sure if would be better to move war downloads to a dedicated task entrypoint?

anyway I don't like the repository_api_explorer_* variable defined here while we are fetching it as a "war download", it seems we are giving it two different levels of importance

gionn (approved):

PR #1091, opened 2025-02-28
OPSEXP-3057 Refactor repository role to use default vars

pmacius:
ref: OPSEXP-3057


roles/repository/defaults/main.yml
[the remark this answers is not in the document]
pmacius: I was thinking to do this in next PR

roles/audit_storage/molecule/default/converge.yml
```diff
@@ -1,9 +1,11 @@
...
 - name: Converge
   hosts: all
+  vars:
+    repository_acs_version: 23.4.1
   roles:
     - role: activemq
     - role: elasticsearch
       vars:
-        elasticsearch_major_version: "{{ '8.x' if acs.version is version('25.0', 'ge') else '7.x' }}"
+        elasticsearch_major_version: "{{ '8.x' if repository_acs_version is version('25.0', 'ge') else '7.x' }}"
```
gionn, line 10: not sure how much is worth having this condition when we are running role tests with a single version

roles/elasticsearch/molecule/default/converge.yml
```diff
@@ -1,9 +1,11 @@
...
   hosts: all
+  vars:
+    repository_acs_version: 23.4.1
   tasks:
     - name: "Include elasticsearch"
       ansible.builtin.include_role:
         name: "elasticsearch"
       vars:
-        elasticsearch_major_version: "{{ '8.x' if acs.version is version('25.0', 'ge') else '7.x' }}"
+        elasticsearch_major_version: "{{ '8.x' if repository_acs_version is version('25.0', 'ge') else '7.x' }}"
```
gionn, line 11: same as before

playbooks/acs.yml
```diff
@@ -29,6 +29,14 @@
 - name: Handle secrets
   ansible.builtin.import_playbook: secrets.yml
 
+- name: Check acs edition
```
gionn, line 32: if thinking bigger is allowed, this is the start of the replacement for the common role that we could call "dynamic facts", let's have a dedicated playbook for it with a more generic play name and import it here

you could also validate that acs_play_repository_acs_edition is either Enterprise or Community to avoid unwanted user mistakes

roles/transformers/defaults/main.yml
```diff
@@ -8,6 +8,8 @@ transformers_tengine_environment:
 transformers_ats_keystore: {}
 transformers_include_ghostscript: false
 
+transformers_repository_acs_edition: Enterprise
```
gionn, line 11: to be really pragmatic, both here and in the repo role you can really just have the boolean var

gionn (approved):

alxgomz (commented):
We could import the dynamic-facts playbook before any other pre-run check playbook so vars defined dynamically are available for use in precheck playbook as well. That would make them unusable without dynamic vars 🤷 but would it really be a problem?

pmacius:
@alxgomz Placement of that playbook can be moved when we will tackle common role. For now it doesn't break anything I think

alxgomz (approved):

PR #1104, opened 2025-03-05
OPSEXP-3118: move ES version to version vars

alxgomz:
Use var to pass ES (major) version so can be overridden in a version specific way when needed
Ref: OPSEXP-3118

### DIFF
--- a/playbooks/group_vars/elasticsearch.yml
+++ b/playbooks/group_vars/elasticsearch.yml
@@ -1,0 +1,1 @@
+acs_play_elasticsearch_major_version: 7.x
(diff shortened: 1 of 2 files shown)


playbooks/group_vars/elasticsearch.yml
```diff
@@ -0,0 +1 @@
+acs_play_elasticsearch_major_version: 7.x
```
gionn, line 1: once we have 25 out, should this be moved to version-dependant vars or not?

alxgomz: only for acs25. Other versions would default here

gionn (approved):

gionn (approved):

PR #1106, opened 2025-03-06
OPSEXP-3057 Move leftover repository group vars into role defaults and playbook group vars

pmacius:
Ref: OPSEXP-3057


playbooks/acs.yml
```diff
@@ -80,8 +80,8 @@
         name: "{{ pg_role }}"
         tasks_from: setup_db
       vars:
-        postgres_db_name: "{{ repo_db_name }}"
-        postgres_db_username: "{{ repo_db_username }}"
+        postgres_db_name: "{{ acs_play_repo_db_name }}"
+        postgres_db_username: "{{ acs_play_repo_db_username }}"
         postgres_db_password: "{{ repo_db_password }}"
```
pmacius, line 85: Would like to move `repo_db_password` in separate PR. Possibly leave for now and maybe create a task to handle secrets?

alxgomz: good point. We should anyway use a different prefix for secrets var names

playbooks/group_vars/trusted_resource_consumers.yml
```diff
@@ -0,0 +1 @@
+acs_play_repository_trusted_urls: "{{ groups['trusted_resource_consumers'] | map('extract', hostvars) | map(attribute='known_urls', default=[]) | flatten }}"
```
alxgomz, line 1: strictly speaking this var is not specific to repo (even though share and repo are part of the same role...) I would remove the `_repository` pattern.

group_vars/repository.yml
alxgomz, line 1: I find it strange there's no change to molecule `group_vars` location after we move to playbook/group_vars
https://github.com/Alfresco/alfresco-ansible-deployment/blob/e01a5eacc79c8dc884edd3f2c04064923e74a198/roles/repository/molecule/default/molecule.yml#L41

Is it working "by accident" or did you expected that (or is it not needed now?)?

alxgomz (changes requested):

group_vars/repository.yml
pmacius: As discussed in DMs we are no changing the path to group_vars because in molecules for roles we are relying on defaults. We'll anyway clean that up when we reach the stage where we make roles really self contained

alxgomz (approved):

PR #1110, opened 2025-03-10
OPSEXP-3040 Remove urls related vars from group vars

pmacius:
Ref: OPSEXP-3040


playbooks/group_vars/all.yml
```diff
@@ -1,3 +1,9 @@
+# REQUIRED: add at least the Share url
+# The first entry will also be used to assume the base URL of this alfresco install
+# For more information please have a look at the
+# [security_doc](https://github.com/Alfresco/alfresco-ansible-deployment/blob/master/docs/SECURITY.md#specify-trustworthy-applications)
+known_urls: []
```
gionn, line 5: should add acs_play_ prefix as well?

group_vars/all.yml
```diff
@@ -1,11 +1 @@
-# REQUIRED: add at least the Share url
-# The first entry will also be used to assume the base URL of this alfresco install
-# For more information please have a look at the
-# [security_doc](https://github.com/Alfresco/alfresco-ansible-deployment/blob/master/docs/SECURITY.md#specify-trustworthy-applications)
-known_urls: []
 acs_play_major_version: 23
```
gionn, line 1: should be safe to move to playbooks/group_vars/all.yaml?

I would leave this file empty with just a comment which tell everything as been moved elsewhere

pmacius: I would like this to be a part of OPSEXP-3162

alxgomz (approved):
Not a fan of spreading the ancient `use_ssl` and `fqdn_alfresco` stuff to more roles, bt I guess we don't have much better options right now

gionn (approved):

PR #1114, opened 2025-03-10
OPSEXP-3106: cleanup ats mtls keystores vars

alxgomz:
Ref: OPSEXP-3106


playbooks/acs.yml
```diff
@@ -423,7 +423,7 @@
         name: "../roles/trouter"
       vars:
         java_version: "{{ acs_play_java_version }}"
-        ats_keystore: "{{ trouter_keystore | default({}) }}"
+        trouter_ats_keystore: "{{ trouter_keystore | default({}) }}"
```
pmacius, line 426: Maybe we should rename `trouter_keystore` to `acs_play_trouter_keystore`
Similar thing is done on repo role when we calculate amps list in pre_tasks

playbooks/acs.yml
```diff
@@ -478,7 +478,7 @@
         sfs_zip_checksum_url: "{{ acs_play_sfs_download_zip_sha1_checksum_url }}"
         sfs_zip_username: "{{ nexus_user }}"
         sfs_zip_password: "{{ nexus_password }}"
-        ats_keystore: "{{ sfs_keystore | default({}) }}"
+        sfs_ats_keystore: "{{ sfs_keystore | default({}) }}"
```
pmacius, line 481: Same as for trouter_keystore

roles/sfs/defaults/main.yml
```diff
@@ -5,8 +5,10 @@ sfs_environment:
     - -Xms128m
     - -Xmx512m
     - $JAVA_OPTS
-utils_sfs: []
-ats_keystore: {}
+sfs_utils: []
+sfs_ats_keystore: {}
+sfs_ats_truststore: "{{ java_home }}/lib/security/cacerts"
+sfs_ats_truststore_pass: changeit
```
pmacius, line 11: Not sure maybe worth to add defaults with acs_play and override them in playbook

gionn (approved):

pmacius (commented):
+ rebase

roles/sfs/defaults/main.yml
alxgomz: I don't know I was mostly sticking to what we add before. Setting these in the playbook would mostly allow users to set a custom path for a truststore.... not sure how much that would be useful

PR #1120, opened 2025-03-12
OPSEXP-3107 Bumps with updatecli v2

gionn:
ref OPSEXP-3107

* Have a single workflow to bump all the artifacts for all the supported versions


.github/updatecli_maven.tpl
pmacius, line 1: Maybe just a bunch of comments or a doc what the parts of this template do. To be easier to understand when we will come back to this after longer time.

pmacius (approved):

PR #1123, opened 2025-03-12
OPSEXP-3169 Support ubuntu 24.04 in vagrant config

gionn:
* Use bento box which also supports arm64
* Use a different box depending on the acs_major_version
* Allow chosing ACS version by setting VAGRANT_ACS_MAJOR_VERSION

OPSEXP-3169


pmacius (approved):
Tested works cool :rocket:

PR #1129, opened 2025-03-17
OPSEXP-3107 Support bumping versions in roles defaults

gionn:
ref OPSEXP-3107


.github/updatecli_maven_23_values.yml
alxgomz, line 1: Can,'t we have a single var for the target file? Roles should always have these keys in the default vars file

PR #1131, opened 2025-03-18
OPSEXP-3174 Prevent http credentials in ansible logs

gionn:
Prevent ansible logging http credentials while using get_url and loop

OPSEXP-3174

### DIFF
--- a/roles/repository/tasks/main.yml
+++ b/roles/repository/tasks/main.yml
@@ -111,6 +111,8 @@
   poll: 0
   tags:
     - molecule-idempotence-notest
+  loop_control:
+    label: "{{ item.url | basename }}"
 
 - name: Download {{ repository_acs_artifact_name }}.zip
   ansible.builtin.get_url:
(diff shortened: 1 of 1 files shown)


alxgomz (commented):
I would also add a `no_log` param to the repsitory role arguments below:
* repository_identity_client_secret
* repository_extra_war_downloads
* repository_amp_downloads
* repository_extra_amp_downloads
As I would expect they also get logged in case of a more verbose play

gionn:
> * As I would expect they also get logged in case of a more verbose play

not going to help as `no_log` lose any effect when ansible run in verbose/debug mode

https://docs.ansible.com/ansible/latest/reference_appendices/logging.html#protecting-sensitive-data-with-no-log

and in general `no_log` make really hard to debug failures, I would avoid using it unless really needed (e.g. executing a shell command with a sensitive argument)

alxgomz (approved):

PR #1133, opened 2025-03-20
OPSEXP-3176 Automate AMI bumps with updatecli

gionn:
OPSEXP-3176


gionn:
> Now shouldn't we schedule it and create PR on bumps?

it even execute ec2 tests automatically now

alxgomz (approved):

PR #1172, opened 2025-05-06
OPSEXP-3233 Fix jdbc driver class not propagated to repo/sync roles

### LINKED ISSUE #1171: Fails with Custom databases other than PostgreSQL
The value of the `repository_jdbc_driver_class` is always `org.postgresql.Driver` unless explicitly overloaded through extra vars, while the documented variable to pass in extra vars to specify a custom DB driver is `acs_play_repo_db_driver`. Same happens for sync service.

In consequence the PostgreSQL driver is always installed, and custom drivers provided through the `configuration_files` directory are not installed. As the `acs_play_repo_db_driver` is used to configure `alfresco-global.properties` the service startup fails because it can not find the DB driver classes.

https://github.com/Alfresco/alfresco-ansible-deployment/blob/c2b3d58f71fb90216ced7d8eb809ba3974806161/roles/repository/tasks/main.yml#L309

We solved this on our side with:

```
repository_jdbc_driver_class: "{{ acs_play_repo_db_driver | default('org.postgresql.Driver') }}"
```
and

```
sync_jdbc_driver_class: "{{ acs_play_sync_db_driver | default('org.postgresql.Driver') }}"
```

gionn:
Fix #1171 OPSEXP-3233


roles/sync/tasks/configure.yml
```diff
@@ -20,7 +20,7 @@
       sql:
         db:
           url: "{{ local_db_url }}"
-          driver: "{{ sync_db_driver }}"
+          driver: "{{ acs_play_sync_jdbc_driver_class }}"
```
pmacius, line 23:
```suggestion
          driver: "{{ sync_jdbc_driver_class }}"
```

jalvarezferr:
Shouldn't group_vars of repository and syncservice set the driver class value based on the values of acs_play_repo_db_driver and acs_play_sync_db_driver respectively?

repository.yml:

```
acs_play_repository_jdbc_driver_class: "{{ acs_play_repo_db_driver | default(default_jdbc_pg_driver_class) }}"
```

syncservice.yml:

```
acs_play_sync_jdbc_driver_class: "{{ acs_play_sync_db_driver | default(default_jdbc_pg_driver_class) }}"
```

Without that, unless we explicitly set acs_play_repository_jdbc_driver_class and acs_play_sync_jdbc_driver_class in extra vars, the role will receive the group_vars values which will be aqain the default (for PostgreSQL). In our previous solution this did not happen as `repository_jdbc_driver_class` was not propagated and we fixed it in the defaults of the role.

The change made to the deployment document, indicating to set acs_play_repository_jdbc_driver_class directly, does not comply with the naming convention of all other external variables. With the change I mention the external variable would remain as it was.

gionn:
what was really confusing me is that both:

* `acs_play_repository_jdbc_driver_class`
* `acs_play_sync_jdbc_driver_class`

are defined as group_vars for their respective groups but never really used.

So I decided to simply get rid of them and make sure the previous `acs_play_repo_db_driver` and `acs_play_sync_db_driver` are properly propagated to their respective role arguments.

PR #1205, opened 2025-06-17
OPSEXP-3267 migrate activemq testinfra tests to ansible

siddavamshi4:
This PR will migrate python module(testinfra) to test ansible roles, testinfra which is no more an available verifier in molecule.
[Removal of testinfra verifier · Issue #3920 https://github.com/https://github.com/ansible/molecule/issues/3920

we will migrate testinfra (python-based) to ansible tasks.
### Related Issue

OPSEXP-3267

### Changes introduced
Removed Python-based Testinfra tests (roles/activemq/molecule/default/tests/test_activemq.py)
Replaced with Ansible assert tasks in the verify.yml


roles/activemq/tasks/main.yml
```diff
@@ -107,7 +107,8 @@
           line: admins={{ activemq_username }}
         - file: jetty-realm
           line: "{{ activemq_username }}: MD5:{{ activemq_password | md5 }}, admin"
-      no_log: true
```
alxgomz, line 110: I know `no_log` is painful but without it `label` on iterator would not prevent from displaying passwords in Ansible output when ran in debug mode for instance

PR #1211, opened 2025-06-19
OPSEXP-3269 Migrate Tomcat testinfra tests to ansible

siddavamshi4:
This PR will migrate python module(testinfra) to test ansible roles, testinfra which is no more an available verifier in molecule.
[Removal of testinfra verifier · Issue #3920 https://github.com/https://github.com/ansible/molecule/issues/3920

we will migrate testinfra (python-based) to ansible tasks.
### Related Issue

OPSEXP-3269

### Changes introduced
Removed Python-based Testinfra tests (roles/tomcat/molecule/default/verify.yml)
Replaced with Ansible assert tasks in the verify.yml


pmacius (approved):

roles/tomcat/molecule/default/verify.yml
```diff
@@ -10,47 +10,45 @@
...
-    changed_when: false
     failed_when: "'/opt/apache-tomcat-' ~ tomcat_version not in tomcat_home_output.stdout"
-  - name: Run tomcat.sh configtest to inspect Catalina environment
-    command: "{{ binaries_folder }}/tomcat.sh configtest"
+
+  - name: To fetch Catalina environment
+    ansible.builtin.command: "{{ binaries_folder }}/tomcat.sh configtest"
     register: catalina_env_output
-    changed_when: false
-    become: true
```
gionn: why did you remove `become`? should not be needed as there is a `become` set in the top play but somehow is triggering kics linter

siddavamshi4: Should i add it back at task level and push ?

PR #1223, opened 2025-07-03
OPSEXP-3205 Handle extra amps in upgrade checks

pmacius:
ref: OPSEXP-3205

### DIFF
--- a/playbooks/acs.yml
+++ b/playbooks/acs.yml
@@ -402,6 +402,17 @@
           }) }}
       when: acs_is_enterprise
 
+    - name: Add extra amps if provided and required properties are set
+      ansible.builtin.set_fact:
+        amps: >-
+          {{ amps | combine({
+            item.name: {
+              'version': item.version
+            }
+          }) }}
+      loop: "{{ acs_play_repository_extra_amp_downloads | default([]) }}"
+      when: acs_play_repository_extra_amp_downloads is defined and item.version is defined and item.name is defined
+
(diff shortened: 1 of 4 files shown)


playbooks/acs.yml
```diff
@@ -402,6 +402,17 @@
           }) }}
       when: acs_is_enterprise
 
+    - name: Add extra amps if provided and required properties are set
+      ansible.builtin.set_fact:
+        amps: >-
+          {{ amps | combine({
+            item.name: {
+              'version': item.version
```
pmacius, line 410: We dont need the repository like in other amps - we only assert the version in prerun upgrade checks

playbooks/acs.yml
```diff
@@ -402,6 +402,17 @@
...
+            }
+          }) }}
+      loop: "{{ acs_play_repository_extra_amp_downloads | default([]) }}"
+      when: acs_play_repository_extra_amp_downloads is defined and item.version is defined and item.name is defined
```
pmacius, line 414:
```suggestion
      when: item.version is defined and item.name is defined
```
I think its not needed

alxgomz (changes requested):
Please add the additionnal attributes of `acs_play_repository_extra_amp_downloads` in. https://github.com/Alfresco/alfresco-ansible-deployment/blob/bba1d0e02dbbb07d9386ba545863d8d8fd4a14b4/docs/deployment-guide.md?plain=1#L688

PR #1229, opened 2025-07-18
OPSEXP-3170 Add workflow to cleanup AWS MQ configuration leftovers

siddavamshi4:
### Related Issue
OPSEXP-3170 
### Summary
This PR introduces a GitHub Action to automatically clean up AWS MQ configurations that:
- Have the name prefix `molecule-`
- Are older than 30 days based on their creation timestamp
- Are not currently attached to any active brokers
- Schedules to run on 1st of every month

#### Note: 
We are using **.LatestRevision.Created** field to check the age and take delete accordingly.

### DIFF
--- a/.github/workflows/aws-mq-cleanup.yml
+++ b/.github/workflows/aws-mq-cleanup.yml
@@ -2,6 +2,11 @@ name: Cleanup MQ Configurations
 
 on:
   workflow_dispatch:
+    inputs:
+      dry_run:
+        type: boolean
+        default: true
+        description: "If true, skips deletion for safe testing"
   schedule:
     - cron: '23 5 1 * *'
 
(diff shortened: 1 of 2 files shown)


.github/workflows/aws-mq-cleanup.yml
```diff
@@ -0,0 +1,81 @@
...
+                --output text)
+
+              if [ -n "$attached" ]; then
+                echo "In use: $name ($id) — broker: $attached"
+                continue
+              fi
+
+              if [ "$created_epoch" -lt "$threshold_epoch" ]; then
+                echo "Deleting: $name ($id) — $age_days days old"
+                aws mq delete-configuration --configuration-id "$id"
```
gionn, line 73: to simplify testing, let's add a `dry-run` input to the workflow dispatch which defaults to true, and then evaluate a variable for this specific step like:

```
env:
  DRY_RUN: ${{ inputs.dry_run && github.event_name != 'schedule' }}
```

and then use it to not run `delete-configuration`:

```
[ "$DRY_RUN" == "false" ] && aws mq...
```

gionn (approved):

PR #1370, opened 2026-03-18
OPSEXP-3923 Bump Rocky and RHEL to 9.7 

pmacius:
OPSEXP-3923


.github/workflows/community.yml
```diff
@@ -97,6 +97,13 @@ jobs:
         with:
           cache-name: community
 
+      - name: Disable unix-chkpwd AppArmor profile
```
pmacius, line 100: https://www.tunbury.org/2025/05/13/ubuntu-apparmor/

.github/workflows/community.yml
```diff
@@ -97,6 +97,13 @@ jobs:
...
+        run: |
+          if sudo aa-status --enabled 2>/dev/null; then
+            sudo ln -sf /etc/apparmor.d/unix-chkpwd /etc/apparmor.d/disable/
+            sudo apparmor_parser -R /etc/apparmor.d/unix-chkpwd 2>/dev/null || true
```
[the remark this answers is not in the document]
pmacius: `/etc/apparmor.d/disable/` missing — This directory is part of the `apparmor` package. The `aa-status --enabled` check ensures we only enter the block if AppArmor is installed and active, which means the package is present and so is this directory. On GitHub Actions Ubuntu runners this is always the case.

`/etc/apparmor.d/unix-chkpwd` not present — `ln -sf` creates symlinks regardless of whether the source file exists (it would just create a dangling symlink, which is harmless). And `apparmor_parser -R` already has `|| true` so it won't fail either.

vars/acs23.yml
```diff
@@ -1,13 +1,15 @@
 supported_os:
   RedHat:
     versions:
+      - 9.7
       - 9.5
```
gionn, line 6: let's add 9.6 for consistency even if it's not explicitly tested

gionn (approved):

PR #1423, opened 2026-06-22
OPSEXP-4157 Move nginx Alfresco logs to /var/log/nginx to satisfy SELinux and out-of-box rotation

gionn:
The reverse-proxy vhosts wrote their access/error logs under /var/log/alfresco, which the common role labels with the SELinux type var_log_t. nginx runs in the httpd_t domain, which is only allowed to open httpd_log_t files, so on SELinux-enabled nodes nginx is denied (avc: denied { open }) and fails to (re)start. The nginx master process opens the logs as root, so the standard permissions without selinux are quitly bypassed.

Point the logs at /var/log/nginx instead, which is httpd_log_t by default and the only location nginx is allowed to write. The paths are now defined once via nginx_access_log / nginx_error_log to avoid duplication across the two vhost templates, and the redundant "nginx." filename prefix is dropped. Files still end in .log so they remain covered by the nginx package's logrotate.

OPSEXP-4157

### DIFF
--- a/roles/nginx/defaults/main.yml
+++ b/roles/nginx/defaults/main.yml
@@ -8,6 +8,12 @@ nginx_vhosts:
 nginx_use_ssl: false
 nginx_fqdn_alfresco: ""
 
+# On SELinux-enabled nodes nginx (httpd_t) may only write logs labelled
+# httpd_log_t, so /var/log/nginx is the only allowed location.
+nginx_logs_folder: /var/log/nginx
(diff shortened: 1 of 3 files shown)


pmacius (approved):

pmacius (commented):
https://github.com/Alfresco/alfresco-ansible-deployment/blob/master/roles/nginx/meta/argument_specs.yml 
Can you update?

pmacius (approved):

PR #1434, opened 2026-07-15
OPSEXP-4180 Rename inventory groups to indexers / legacy_indexers with backward-compatible aliases

pmacius:
## What

Introduces `indexers` (Elasticsearch-based indexing) and `legacy_indexers` (Solr-based indexing) as the canonical inventory group names, replacing `search_enterprise` and `search`. The old names are retained as backward-compatible sub-groups, so existing inventories keep working without any changes. The separate `elasticsearch` group (the ES cluster backing Search Enterprise) is unchanged.

## Why

The names `search_enterprise` and `search` convey the product edition rather than the indexing technology, which is confusing and blocks a future goal of serving both Enterprise and Community Elasticsearch deployments from a single group. The new names describe what the hosts actually do and set the stage for that consolidation.

## How

Backward/forward compatibility is achieved at the reference layer, not just the inventory layer, because Ansible cannot force a user's flat inventory (old names only) to define a new parent group, and a pure new-name inventory won't define the old groups.

- **Inventories** — the three shipped inventories use only the canonical `indexers` / `legacy_indexers` groups and no longer mention the deprecated names, since they are only referenced when provisioning a new environment. Backward compatibility for existing inventories is handled entirely in the playbook/role code (see below), not by defining alias groups in the inventory. A harmless `Could not match supplied host pattern` warning is emitted for whichever group name is absent from a given inventory; this is expected given the union patterns.
- **`hosts:` play targets** — rewritten to union patterns (e.g. `indexers:search_enterprise`, `legacy_indexers:search`) so a play matches hosts under either the new or the old name.
- **`groups[...]` lookups** — playbooks use two new helper vars (`indexer_hosts` / `legacy_indexer_hosts`, defined in `playbooks/group_vars/all.yml`) that union the canonical group with its deprecated alias; role-internal files (`roles/common/defaults/main.yml`, `alfresco-global.properties.j2`) inline the same union so isolated role molecule runs stay self-contained.
- **group_vars** — the static artifact-coordinate vars previously in `search.yml` / `search_enterprise.yml` are moved into `all.yml` (global scope) so they resolve regardless of which group a host lives in; without this, a pure new-name inventory would leave those vars undefined. `updatecli` is repointed to the new file location (JSONPath keys unchanged).
- **Reindex play** — `search-enterprise-reindex.yml` uses `run_once` instead of the `search_enterprise[0]` subscript, because a subscript does not reliably select the first host of a union pattern (verified: `indexers:search_enterprise[0]` ran on all hosts). An `.ansible-lint-ignore` entry is added since the `run-once[task]` rule only matters under `strategy: free`, which this play does not use.
- **Molecule** — the four dedicated search scenarios (`roles/search`, `roles/search_enterprise`, `opensearch`, `multimachine`) are flipped to the new group names to exercise them in CI; the incidental scenarios keep the old names and continue to pass via the dual resolution above.
- **Docs** — deprecation note added to the upgrade guide; inventory group list and the Search Services example updated.

## Verification

- `ansible-inventory --graph` resolves correctly for the shipped inventories, a pure new-name inventory, and an old flat inventory.
- The helper vars and the relocated `acs_play_search_enterprise_*` vars resolve for a host placed directly in `indexers`, confirming a pure new-name inventory works identically.
- `--syntax-check` passes on all touched playbooks; `yamllint` and `ansible-lint` are clean (0 failures).

OPSEXP-4180

### DIFF
--- a/inventory_local.yml
+++ b/inventory_local.yml
@@ -16,21 +16,17 @@ all:
       children:
         repository:
 
-    # Elasticsearch-based indexing hosts (canonical group, default search engine -
+    # Elasticsearch-based indexing hosts (default search engine -
     # not available in Community edition)
     indexers:
       children:
         repository:
-        # Deprecated alias retained for backward compatibility (e.g. dynamic/tag-based
-        # inventories still populating it); see docs/playbook-upgrade.md
-        search_enterprise:
 
     # Solr-based indexing hosts (legacy engine; alternative to Enterprise Search and the
     # only available option for the Community edition)
     legacy_indexers:
       children:
         # repository:
-        search:
 
     elasticsearch:
       children:
--- a/inventory_local.yml
+++ b/inventory_local.yml
@@ -20,18 +20,17 @@ all:
     # not available in Community edition)
     indexers:
       children:
-        # Deprecated alias retained for backward compatibility; see docs/playbook-upgrade.md
+        repository:
+        # Deprecated alias retained for backward compatibility (e.g. dynamic/tag-based
+        # inventories still populating it); see docs/playbook-upgrade.md
         search_enterprise:
-          children:
-            repository:
 
     # Solr-based indexing hosts (legacy engine; alternative to Enterprise Search and the
     # only available option for the Community edition)
     legacy_indexers:
       children:
+        # repository:
         search:
-          children:
-            # repository:
 
     elasticsearch:
       children:
--- a/inventory_local.yml
+++ b/inventory_local.yml
@@ -16,15 +16,23 @@ all:
       children:
         repository:
 
-    # Solr search engine (an alternative to Enterprise Search, and the only available option for the Community edition)
-    search:
+    # Elasticsearch-based indexing hosts (canonical group, default search engine -
+    # not available in Community edition)
+    indexers:
       children:
-        # repository:
+        # Deprecated alias retained for backward compatibility; see docs/playbook-upgrade.md
+        search_enterprise:
+          children:
+            repository:
 
-    # Enterprise Search (default search engine - not available in Community edition)
-    search_enterprise:
+    # Solr-based indexing hosts (legacy engine; alternative to Enterprise Search and the
+    # only available option for the Community edition)
+    legacy_indexers:
       children:
-        repository:
+        search:
+          children:
+            # repository:
+
     elasticsearch:
       children:
         repository:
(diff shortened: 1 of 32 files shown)


playbooks/group_vars/search.yml
pmacius, line 1: Not sure about it maybe for now we move to all and after deprecating we move back to specific group vars

inventory_local.yml
```diff
@@ -16,15 +16,22 @@ all:
       children:
         repository:
 
-    # Solr search engine (an alternative to Enterprise Search, and the only available option for the Community edition)
-    search:
+    # Elasticsearch-based indexing hosts (canonical group, default search engine -
+    # not available in Community edition)
```
gionn: it's triggering me a bit but until we introduce the new role then it's true

vars/acs23.yml
gionn, line 1: revert updatecli changes

inventory_ssh.yml
```diff
@@ -27,17 +27,26 @@ all:
...
 
-    # Solr search engine (an alternative to Enterprise Search, and the only available option for the Community edition)
-    search:
+    # Elasticsearch-based indexing hosts (canonical group, default search engine -
+    # not available in Community edition)
+    indexers:
+      # Deprecated alias retained for backward compatibility (e.g. dynamic/tag-based
+      # inventories still populating it); see docs/playbook-upgrade.md
+      children:
+        search_enterprise:
```
gionn: I don't think we need to mention deprecated groups in all the new inventories, users will look at these only when provisioning a new env, so retrocompat should really be handled within the ansible code (as we are already doing)

playbooks/prerun-checks.yml
```diff
@@ -29,14 +29,14 @@
...
+          - indexer_hosts | length == 0
         fail_msg: mTLS enabled but not yet supported for search_enterprise
       when: ats_mtls_required and ats_mtls_capable
 
     - name: Assert that search-enterprise group is empty when edition is community
       ansible.builtin.assert:
         that:
-          - groups.search_enterprise | default([]) | length == 0
+          - indexer_hosts | length == 0
         fail_msg: >-
```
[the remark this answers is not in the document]
gionn: check

playbooks/group_vars/search.yml
gionn: if the alternative is to duplicate vars then yes better keep as you did

inventory_local.yml
pmacius: I will remember about that

gionn (approved):

PR #1458, opened 2026-08-20
OPSEXP-4247 Wire molecule tests to the group_vars artifact repository

gionn:
## Summary
Molecule scenarios only linked host_vars, so the repository URL variables defined in playbooks/group_vars weren't reachable from role tests, and some converge files hardcoded artifacts.alfresco.com directly. This links group_vars the same way in every role's molecule.yml, and every role that downloads an artifact now overrides its repository var with the corresponding acs_play_* variable (the same one playbooks/acs.yml passes in production), so changing artifacts_repositories in playbooks/group_vars/all.yml propagates to every molecule test without touching each role individually. Role defaults keep their artifacts.alfresco.com fallback for standalone role usage — only the test wiring changes.

## Test plan
- [ ] yamllint passes on all changed files
- [ ] Run molecule test for sync role and confirm it resolves the repository/zip URLs from playbooks/group_vars
- [ ] Run molecule test for adf_app role and confirm it resolves acc_repository/adw_repository
- [ ] Run molecule test for repository, search, search_enterprise, sfs, trouter and transformers roles and confirm they resolve their artifact repository from playbooks/group_vars
- [ ] Point artifacts_repositories.base_url at an alternate host and confirm the change propagates to all of the above without further edits

OPSEXP-4247

### DIFF
--- a/roles/activemq/molecule/default/molecule.yml
+++ b/roles/activemq/molecule/default/molecule.yml
@@ -34,3 +34,4 @@ provisioner:
   inventory:
     links:
       host_vars: host_vars
+      group_vars: ../../../../playbooks/group_vars
--- a/roles/adf_app/molecule/default/converge.yml
+++ b/roles/adf_app/molecule/default/converge.yml
@@ -4,10 +4,10 @@
   vars:
     acc_v: 10.3.0
     acc_id: alfresco-control-center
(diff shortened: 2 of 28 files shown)


pmacius (approved):
LGTM

PR #1466, opened 2026-09-04
OPSEXP-4339 Bump versions for oxygen release

pmacius:
Bumps the component versions deployed by the playbook for the Oxygen release, across all three supported ACS lines. The ACS 25 line moves from 25.4 to 25.5, so the updatecli AMP release branch and the API Explorer version follow it.

## Version bumps

Common to 26.2, 25.5 and 23.7:

| Component | From | To |
|---|---|---|
| Audit Storage | 1.3.3 | 1.3.4 |
| Search Services | 2.0.20 / 2.0.21 | 2.0.22 |
| Search Enterprise | 5.7.0 | 5.8.0 |
| Shared File Store | 4.4.3 | 4.4.4 |
| Sync Service | 5.3.5 | 5.3.6 |
| Transform Router | 4.4.3 | 4.4.4 |
| All-In-One T-Engine | 5.4.3 | 5.4.4 |
| PDF Renderer | 1.3.0-81 | 1.3.0-85 |

Per line:

| Component | 26.2 | 25.5 | 23.7 |
|---|---|---|---|
| Repository / Share | 26.2.0 → 26.2.2 | 25.4.1 → 25.5.0 | 23.7.1 → 23.7.3 |
| API Explorer | — | 25.4.0 → 25.5.0 | — |
| OpenJDK | 21.0.11+10 → 21.0.12+8 | 17.0.19+10 → 17.0.20+8 | 17.0.19+10 → 17.0.20+8 |
| Apache Tomcat | 11.0.22 → 11.0.25 | 10.1.55 → 10.1.59 | 10.1.55 → 10.1.59 |
| Apache ActiveMQ | 6.2.7 → 6.2.9 | — | — |
| Google Docs AMP | — | 4.1.1 → 5.0.0 | — |
| AOS AMP | — | 3.4.1 → 3.4.2 | — |
| Device Sync AMP | — | 5.3.5 → 5.3.6 | — |

The Community repository on 26.2 stays at 26.2.0, and Community search stays at 5.7.0.

## Docs

The [versions table in the overview](docs/overview.md#versioning) is updated to match the new values, including the `25.4` → `25.5` column header. The Repository and Share rows now read `26.2.2 (26.2.0)` because Enterprise and Community diverge on that line for the first time, using the table's existing `(Community)` parenthesis convention.

## Molecule test fixes

Two `verify.yml` assertions were pinned to values this bump changes.

**`repository`** — the scenario asserts the deployed repository version, hardcoded to `26.2.0`; bumped to `26.2.2`.

**`search_enterprise`** — Elasticsearch connector 5.8.0 renamed the log line the scenario greps for:

```
5.7.0: Refreshing accepted content media-type cache, accepted type count: 70
5.8.0: Refreshed media-type cache from 1 endpoint(s), accepted type count: 70
```

Converge and idempotence pass cleanly, so this was the assertion failing and not the deployment. The check is the regression guard added in 8f11ace5 (OPSEXP-1813) for the `ALFRESCO_ACCEPTED_CONTENT_MEDIA_TYPES_CACHE_BASE_URL` routing: it proves the connector reached the transform-config endpoint and got a healthy media-type list back. The count and its threshold are unchanged, so the guard still holds — only the wording moved. The pattern now anchors on the stable part of the message and matches both wordings:

```yaml
log_expected_regex: |-
  media-type cache.*accepted type count: ([7-9]\d|[1-9]\d{2,})$
```

Two related fixes in the same assertion, both flagged as warnings in the job log:

* `multiline=True` — the `$` anchor was evaluated without `re.M`, so it only ever matched because the media-type line happened to be the last line of `journalctl` output. Any log line arriving after it would have failed the test spuriously.
* `is ansible.builtin.regex(...)` in place of `| regex_search(...)` — `regex_search` returns `None` rather than a boolean, which raised `DEPRECATION WARNING: Conditional result ... was of type 'NoneType' ... will be removed in version 2.19`, plus a warning about Jinja delimiters inside a conditional. The `regex` test returns a real boolean and takes the variable directly, clearing both. It is the same test already used further down this file.

OPSEXP-4339


roles/search_enterprise/molecule/default/verify.yml
```diff
@@ -26,11 +26,11 @@
...
       vars:
         log_expected_regex: |-
-          Refreshing accepted content media-type cache, accepted type count: ([7-9]\d|[1-9]\d{2,})$
+          media-type cache.*accepted type count: ([7-9]\d|[1-9]\d{2,})$
       ansible.builtin.assert:
         that:
-          - not error_log.stdout | regex_search(' ERROR ')
-          - error_log.stdout | regex_search('{{ log_expected_regex }}')
+          - error_log.stdout is not ansible.builtin.regex(' ERROR ')
+          - error_log.stdout is ansible.builtin.regex(log_expected_regex, multiline=True)
```
gionn, line 33: what triggered these changes

vars/acs25.yml
```diff
@@ -25,26 +25,26 @@ supported_os:
...
-acs_play_search_version: 2.0.21
+acs_play_search_version: 2.0.22
 acs_play_java_version: "{{ acs_play_default_java_version }}"
-acs_play_search_enterprise_version: 5.7.0
-acs_play_sfs_version: 4.4.3
-acs_play_sync_version: 5.3.5
+acs_play_search_enterprise_version: 5.8.0
+acs_play_sfs_version: 4.4.4
+acs_play_sync_version: 5.3.6
 acs_play_cic_connector_version: 1.0.1
```
gionn: cic connector is lacking updatecli support?

pmacius: Added in https://github.com/Alfresco/alfresco-ansible-deployment/pull/1469

gionn (approved):

